Issue #13, opened 2026-03-26
Add In Place Istio Upgrade Example.


utkarpun:
Add an in place istio upgrade example using the [Istio on EKS Samples](https://github.com/aws-samples/istio-on-eks).  

Also add Version Support Matrix for Istio and Karpenter in eks-upgrader reference files.

PR #16, opened 2026-03-26
Creating foundation for istio upgrade example

utkarpun:
*Issue #, if available:* #13 

*Description of changes:* 
- Create an example to show how eks-upgrader skill can help deal with custom addons like Istio while upgrading

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


utkarpun:
Parking this PR for now: the eks-upgrader skill it depends on is being moved to experimental/eks-upgrader while we iterate on a smaller scoped upgrade story (PR #36, eks-upgrade-check). Will revisit this Istio example once the upgrader stabilizes — likely against a future v2 of the skill rather than the current shape.

PR #107, opened 2026-07-05
docs(eks-genai): clarify Karpenter consumes (never purchases) Capacity Blocks/ODCRs

Hantzley:
## What

Expands the **EC2 Capacity Blocks for ML** section in `skills/eks-genai/references/cluster-and-scheduling.md` to explicitly clarify that **Karpenter only *consumes* pre-existing capacity reservations — it never *purchases* them.**

## Why

The current section covers the wiring mechanism (`capacity-type: reserved` + `capacityReservationSelectorTerms`) and says to "book in advance," but never states that Karpenter does not create/buy a Capacity Block. Field SAs repeatedly assume Karpenter will auto-provision a CB the same way it calls `CreateFleet` to launch an EC2 instance. This closes that gap.

## What's added

- **"Karpenter consumes reservations — it never purchases them"** subsection: the `CreateFleet` "launches ≠ acquires" framing.
- **Manual purchase workflow**: `describe-capacity-block-offerings` → `purchase-capacity-block`, `cb-…` offering ID vs `cr-…` reservation ID.
- **Purchase constraints**: 1-day increments up to 14 days / 7-day up to 182; 64 instances/block (256 across blocks); upfront payment; `payment-pending` → `scheduled`; no-cancel.
- **Karpenter wiring + prioritization**: `capacityReservationSelectorTerms`, `reserved` type, $0 modeling, fallback.
- **Version gates**: ODCR v1.3, Capacity Blocks v1.6, interruptible v1.10, `ReservedCapacity` feature gate.
- **CB reclamation timing**: EC2 terminates 30/60 min before end; Karpenter drains 10 min prior; `do-not-disrupt` caveat.
- **EKS Auto Mode nuance**: open-ODCR auto-matching vs explicit CB selection.
- **6 authoritative source links** added to Sources.

## Verification

Verified 2026-07-05 against the Karpenter ODCR docs, EC2 Capacity Blocks docs (purchase + pricing/billing), and EKS Auto Mode / ml-node-pools docs. Markdown-only change; no code paths affected.


utkarpun (dismissed):
## 8-Lens Review: CHANGES REQUESTED

---

### CI Blocker

The `docs auto-gen blocks are in sync` check is failing. Run `./misc/update-pages.sh` and commit the regenerated output.

---

### IMPRECISE (must fix)

| # | Issue | Fix |
|---|-------|-----|
| 1 | **"once the block is `scheduled`/active"** conflates two distinct states. A CB in `scheduled` state has zero available capacity (AWS docs: "Before a Capacity Block reservation begins, the available capacity appears as zero"). Karpenter can only launch into a CB once it transitions to `active`. This could lead to incorrect advice that pods will be fulfilled before the CB window opens. | Change to "once the block is **active**" |
| 2 | **"CBs can't be cancelled once reserved"** is factually true but materially incomplete. CB Extensions exist: unlimited extensions, same duration increments (1-day up to 14 days, 7-day up to 182 days total), paid upfront per extension. Without this, an SA would advise customers there is zero flexibility when training overruns, missing the primary operational escape hatch. | Add after the no-cancel bullet: "Extensions ARE possible (request before expiry; not guaranteed, capacity-dependent). See [Extend Capacity Blocks](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-blocks-extend.html)." |

---

### Architecture concern: hardcoded operational detail

The new subsection is ~48 lines. The evergreen behavioral guidance is:
- Karpenter consumes, never purchases (the core insight)
- Manual-first, Karpenter-second workflow shape
- The Auto Mode ODCR blast radius warning

The rest hardcodes operational detail that will go stale:

| Hardcoded value | Staleness risk |
|---|---|
| `v1.3`, `v1.6`, `v1.10` version gates | Changes when features promote to stable |
| `ReservedCapacity` feature gate name | Could be removed when stable |
| `64 instances/block (256 across blocks)` | Service limit, AWS increases these |
| `30 min` / `60 min` / `10 min` timing values | Could change with service updates |
| `5 min-12 hrs` payment clearing window | Operational detail |
| Duration increment rules | Could expand |
| `2026-07-14T00:00:00Z` in CLI example | Already stale |

**Recommendation:** Slim the section to behavioral guidance + a brief purchase workflow framing (without reproducing limits and timings that live in AWS docs). The 6 source links you added are excellent for this. Link to them for operational detail rather than inlining it.

Consider adding a guard statement in SKILL.md to address the misconception at the steering level:
```
NEVER suggest Karpenter can purchase or create a Capacity Block. It only consumes pre-existing reservations.
```
This ensures the model sees the correction whenever it advises on GenAI scheduling, regardless of whether the reference doc is loaded.

---

### NUANCED (should fix)

| # | Issue | Fix |
|---|-------|-----|
| 3 | **Auto Mode ODCR blast radius** is correctly stated but has no remediation. Configuring CB selector terms on one NodeClass silently breaks open-ODCR consumption on ALL other NodeClasses. | Add: "Once opted in, add explicit ODCR selector terms to all other NodeClasses that should continue using reservations." |

---

### Advisory (post-merge, not blocking)

1. Instance type list says "p5/p5e/trn1/trn2" but supported list now includes p4d/p4de/p5en/p6-b200/p6-b300 (pre-existing text, not introduced by this PR)
2. IAM: `ec2:PurchaseCapacityBlock` is a spending-authority permission. One-liner noting it should be restricted to FinOps/platform roles would be useful.
3. Single-AZ constraint is implicit via EFA/UltraCluster context but worth a note.

---

**Summary:** The core insight (Karpenter consumes, never purchases) is correct and fills a real gap. Fix the `scheduled`/`active` conflation, add the extensions escape hatch, slim hardcoded details in favor of doc links, and consider a SKILL.md guard for the misconception. CI needs `update-pages.sh`.

Issue #126, opened 2026-07-10
eks-ingress-migration: zero-ingress estate rated RED-5 instead of TRIVIAL


utkarpun:
## What

`skills/eks-ingress-migration/references/ingress-discovery.md` line 26 rates "No controller found" as 🔴 5 (High):

```
- 🔴 5 (High): No controller found, or controller pods in CrashLoopBackOff
```

## Why this is wrong

This skill assesses **migration difficulty**. An empty ingress estate (no controller, no routes) means there is *nothing to migrate* — the correct rating is TRIVIAL / Impact 1, not High. CrashLoopBackOff is a separate concern (broken controller ≠ absent controller).

Conflating "absent" with "broken" produces absurd scores: a fresh cluster with no ingress installed gets rated harder to migrate than one with a healthy single-controller estate.

## Suggested fix

Split the bullet into two:

```
- 🟡 1 (Low/Trivial): No ingress controller found — nothing to migrate
- 🔴 5 (High): Controller pods in CrashLoopBackOff or unreachable
```

utkarpun:
Fixed by #129 (merged as `3e42a559`). The severity-model rework scores migration difficulty by what's at stake rather than machinery presence, so a zero-ingress / empty estate now short-circuits to 100 / TRIVIAL instead of RED-5 — verified by the eval-3 regression case (empty ⇒ TRIVIAL). Closing; the PR referenced this issue but didn't use a closing keyword, so it stayed open after merge.

Issue #127, opened 2026-07-10
eks-ingress-migration: stale HTML-report directives + Tier-B vs alb-migration impact contradiction


utkarpun:
## What — stale HTML-era directives

Several references still instruct the model to record data "for the 3D visualization" / "in the topology JSON" — artifacts that only exist in the HTML report path (report_to_html.py). The markdown-only report path has no 3D visualization or topology JSON.

**Locations:**
- `references/traffic-routing.md:90` — "Record group membership in the topology JSON"
- `references/traffic-routing.md:94` — "the 3D diagram and Routing Topology under-count them"
- `references/traffic-routing.md:100` — "Record all routing patterns … for the 3D visualization"
- `references/ingress-resources.md:91` — "Record every Ingress→backend mapping for the 3D visualization"

These directives are harmless noise when only markdown is produced, but they waste context and may confuse the model into hallucinating JSON output structures.

## What — Tier-B impact cap contradicts alb-migration.md

`references/report-generation.md:34` (Tier-B scoring rule):
> CORS / IP allowlist / rate-limit … **Never Impact 4–5.**

`references/alb-migration.md:92`:
> Treat CORS / rate-limit / IP-allowlist conversions as **🔴 High impact/effort**, not a simple Medium swap.

These directly contradict each other. report-generation.md caps these at Impact 3 while alb-migration.md instructs rating them High (≥4). The alb-migration rationale is sound (WAF cannot inject CORS response headers, rate rules are coarser), so the Tier-B cap in report-generation.md appears to be stale from an earlier scoring model.

## Suggested fixes

1. **Stale directives:** Replace "topology JSON" / "3D visualization" with "the Routing Topology table" (the markdown artifact that actually exists in every report path).

2. **Impact contradiction:** Either raise the Tier-B cap in report-generation.md to allow Impact 4–5 when CORS/rate-limit requires code changes, or downgrade alb-migration.md to match the cap. The alb-migration rationale is stronger — recommend updating report-generation.md line 34 to: "Rate Impact 2–3 when workaround is config-only; Impact 4 when it requires application code changes (e.g. CORS response headers). Never Impact 5."

kritktongrakroj:
issue remark:

Part 1 (stale HTML directives) — the four discovery directives (traffic-routing.md §5.4/§5.5, ingress-resources.md §6) now
  point at the Routing Topology table instead of "topology JSON / 3D visualization." One correction on the premise: those
[...]

PR #129, opened 2026-07-10
fix(eks-ingress-migration): stop rating an empty/absent ingress

kritktongrakroj:
(#126 Issue Fix)

Rework the severity model so migration difficulty is scored by what is at stake, not by how hard a fix looks or whether machinery merely exists.

Priority order for every Impact rating is now explicit:
  1. business logic / revenue (the live traffic at stake)
  2. security / reputation
  3. effort to remediate -- and effort is NOT a severity driver (the same change is trivial for an expert, hard for a novice).

Three independent dimensions now stack instead of overriding one another: migration-difficulty (config complexity), tech-debt (present-but-broken), and security (CVE/EOL). Each is a separate Score Breakdown row.

Presence vs. absence (the core fix for #126):
  - Absent controller / empty estate = 0 (non-event) -- nothing to migrate.
  - Empty estate short-circuits to 100 / TRIVIAL with a "nothing to migrate" note (cluster/node upgrades are out of scope, not counted as migration).
  - Orphaned Ingress objects with no controller = dead config = 0, plus a loud "Migration Crew Alert" note so it is visible, not hidden.
  - EOL/CVE severity scales with the live business traffic served; 0 when the controller serves nothing or is broken.

Deliberate deviation from the issue's suggested fix ---------------------------------------------------- The issue suggested CrashLoopBackOff stay High (5). This change rates it Low (1) tech debt + a mandatory cleanup note instead. Reasoning, via the bridge-builder model:

  - Absent controller = a road that was never built. No cars travel, there is nothing to migrate, and there is no blast radius. Non-event -> 0.

  - Broken controller (CrashLoopBackOff/unreachable) = a road that was started and left half-built / jammed. It carries no live traffic either, so it is NOT a migration-difficulty item -- but because someone deployed it and walked away, it can mis-align or mis-configure other systems that assume it works. That ignored, half-finished state is tech debt: a small (1) deduction plus a note making the cluster owner responsible to fix or remove it.

  - A true High (5) is reserved for things that actually threaten live business traffic -- e.g. an EOL controller with an active CVE serving revenue-critical routes.

Broken scoring worse than absent (1 vs 0) is therefore intentional: "started and abandoned" has a wider blast radius than "never started." If a broken controller's config is itself complex, that complexity is still scored separately as migration difficulty when it is migrated -- the -1 is only the operational-hygiene signal, not the migration cost.

Scope
-----
The shared "Rating model" header is duplicated across six sibling reference files (traffic-routing, alb-migration, migration-risk, dns-certificates, ingress-resources, migration-plan). All six are updated to the new priority wording so nothing contradicts the model; the detailed stacking / presence rules live in ingress-discovery.md.

Regenerated the Docusaurus mirror under misc/website/docs/ via ./misc/update-pages.sh so the docs-sync CI check stays green.

- [ ] Ran `/apex:new-skill` (or walked the equivalent manual steps in `CONTRIBUTING.md`)
- [ ] `skills/<skill>/SKILL.md` present with valid frontmatter
- [ ] `misc/evals/<skill>/triggering.json` authored (≥16 prompts; balanced positives and near-miss negatives)
- [ ] For each neighbour: `misc/evals/<neighbour>/SIBLING_MAP` gained a bullet and its `triggering.json` gained the matching negatives (via `update_sibling_map.py` or hand-edit)
- [ ] Ran the `update-docs` skill and committed any resulting changes (regenerated wrappers/manifest, marker-block updates, prose edits)


utkarpun (dismissed):
# 9-Lens Review — PR #129 (fix/126-ingress-severity-model): CHANGES_REQUESTED

Verdict: the core fix for #126 is correct and well-motivated — an empty estate no longer scores worse than a healthy one, and the deviation rationale is documented in the files themselves, not just the PR body. But the new "serves no live traffic ⇒ non-event" generalization is factually wrong in two security-relevant ways (both verified against primary sources), and the new rules contradict each other and the skill's own eval in several places. Must-fixes below, then should-fixes.

---

## INCORRECT (must fix)

### M1 — "CVE on zero-route machinery = non-event" is refuted by CVE-2025-1974 (IngressNightmare)
`references/ingress-discovery.md` §1.4, 🟢 0 bullet: *"An unexploitable CVE on machinery that carries no traffic is a non-event; do not deduct (record it as an informational note only)"* — with "zero live routes" listed as a qualifying condition.

CVE-2025-1974 (CVSS 9.8) is exploitable on ingress-nginx `< v1.11.5 / < v1.12.1` via the validating admission webhook, **with zero Ingress resources configured**, from the pod network, with no credentials, leading to cluster-wide secret access. The webhook is enabled by default in Helm installs. A healthy zero-route vulnerable ingress-nginx is a critical security finding; this rule files it as an informational note.

Fix: "zero live routes" must not imply "unexploitable." Scope the non-event bucket to *data-plane traffic exposure*, and state explicitly that control-plane attack surface (admission webhooks) is exposed regardless of routes — a known-CVE controller stays a security finding even with zero routes.

Sources: https://kubernetes.io/blog/2025/03/24/ingress-nginx-cve-2025-1974/ · https://www.wiz.io/blog/ingress-nginx-kubernetes-vulnerabilities

### M2 — "Broken/absent controller ⇒ serves no live traffic" is wrong for the AWS Load Balancer Controller
`references/ingress-discovery.md` §1.1 tech-debt bullet ("it serves no live traffic") and §1.4 🟢 0 bullet ("absent, broken … = non-event"). The AWS LBC is control-plane only: the data plane is the ALB/NLB, which keeps forwarding traffic to registered targets while the controller pod crashloops — and an *uninstalled* controller leaves previously-provisioned ALBs running and serving (deletion requires the controller to process the `ingress.k8s.aws/resources` finalizer; AWS docs explicitly warn the ALB remains). Under the current wording, a live internet-facing estate behind a broken/removed LBC scores 0 on the security dimension.

Also true for ingress-nginx in the partial-failure case: with multiple replicas, "pods in CrashLoopBackOff" can be one replica while the others serve production.

Fix: gate the tech-debt/non-event downgrade on **verified** zero traffic (all replicas down for in-pod data planes like nginx; LB/target-group state for LBC-managed estates), not on pod status alone.

Sources: https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/how-it-works/ · https://docs.aws.amazon.com/eks/latest/userguide/delete-cluster.html

### M3 — Orphaned-config estate triggers two contradictory rules simultaneously
`report-generation.md` §1.0-B says: orphaned Ingress objects + no controller ⇒ everything 0, "if there are no other live findings, the estate scores 100 / TRIVIAL." But `ingress-discovery.md` §1.2 Medium bullet says: *"no IngressClass defined **while Ingress resources exist**"* = Impact 3–4 — which is true in exactly that estate. The same cluster yields 100/TRIVIAL and a Medium finding from two rules; the model must violate one of them. (§1.2's 🟢 0 bullet — "No IngressClass defined AND no controller / no Ingress resources" — is also ambiguous about whether "/" is AND or OR and doesn't clearly cover Ingress-present.)

Fix: make §1.2's 0-band explicitly cover the orphaned case (no controller ⇒ IngressClass findings are moot/0 regardless of Ingress presence), or route §1.2 through the §1.0 short-circuit.

### M4 — Broken controller's config: "scored normally" vs "serves no live traffic ⇒ 0"
`report-generation.md` §1.0 distinction note says a broken controller's *"config complexity is scored normally as migration difficulty"*, but the new 🟢 0 Impact band (Impact Indicator table; same in SKILL.md) defines non-event as anything that "serves no live traffic" — and §1.1 states a broken controller serves no live traffic. A CrashLooping nginx with a `configuration-snippet` route is Tier-A Impact 5 under one reading and non-event 0 under the other; both are rule-compliant. Also inconsistent with §1.0-B, where identical dead routes with *no* controller are 0.

Fix: define the non-event band by estate state (absent/orphaned), not by the "serves no live traffic" predicate, and state explicitly whether a broken controller's routes count as migratable config (they presumably do — the migration will resurrect them).

### M5 — The verbatim Migration Crew Alert can assert a false score
`report-generation.md` §1.0-B mandates emitting the alert **verbatim**, and the block hardcodes *"Scored as 100 / TRIVIAL."* But 100 is conditional on "no other live findings." In a mixed estate (orphaned nginx objects + a live EOL controller serving traffic scoring, say, 66/HARD), the mandatory note still prints "100 / TRIVIAL", contradicting the headline score and the "Total must equal 100 − score" audit rule. The block also hardcodes "NGINX" though orphans may belong to any controller class.

Fix: parameterize the note (score placeholder + controller class), or restrict "verbatim" to the empty-estate-only case.

### M6 — The skill's own eval now fails compliant output
`misc/evals/eks-ingress-migration/evals.json` eval 1 expectation: *"No GREEN/AMBER/RED rating language is used — only Impact 1-5."* The PR makes every report render a 🟢 0 Non-event band (Impact Indicator table is mandatory, and §1.6 requires listing non-events at 0 pts). A grader holding the report to "only Impact 1-5" fails correct output.

Fix: update the expectation to "Impact 0–5" in the same PR. Strongly recommended alongside: add an eval case for the exact #126 regression (empty estate ⇒ 100/TRIVIAL; orphaned config ⇒ alert emitted) so the fixed behavior is tested.

### M7 — An untouched line still makes effort a score input
`references/report-generation.md`, Assessment Summary section: *"Impact = how hard/risky it is to transfer or replace that feature … **plus the effort to change**."* This directly contradicts the PR's core rule ("Effort is NOT a severity driver"), stated ~10 times elsewhere in the same files. The model receives conflicting instructions for the Assessment Summary section specifically.

Fix: reword to match the new model.

### M8 — Both new IngressClass claims misstate Kubernetes behavior
`references/ingress-discovery.md` §1.2:
- Medium bullet: classless Ingresses *"fall back to an ambiguous/implicit default"* — there is no implicit default. Spec-compliant controllers **ignore** Ingresses without a class unless explicitly opted in (ingress-nginx `--watch-ingress-without-class`, default false in the Helm chart; AWS LBC requires a defined default IngressClass or the deprecated annotation). The real risk is *silent non-reconciliation*, not ambiguous defaulting.
- High bullet: multiple default IngressClasses ⇒ *"non-deterministic admission"* — admission is deterministic. Since Kubernetes v1.25 the DefaultIngressClass admission plugin **chooses the newest default** (creationTimestamp, name tiebreaker) rather than failing; all EKS-supported versions in 2026 behave this way. (Note: the kubernetes.io docs page still describes the older ≤1.24 rejection behavior — the source of truth is the admission plugin code, kubernetes/kubernetes#110974.)

Fix: reword both mechanisms. Severity levels can stay as chosen, but the stated behaviors must be right — this skill's audience will act on them.

Sources: https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class · https://kubernetes.github.io/ingress-nginx/user-guide/k8s-122-migration/ · https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/ingress_class/ · https://github.com/kubernetes/kubernetes/pull/110974

---

## IMPRECISE / MISLEADING (should fix)

**S1 — Broken controller = Impact 1 conflates "abandoned experiment" with "active outage."** For ingress-nginx the controller pod *is* the data plane: CrashLoopBackOff on all replicas means every bound route is down *right now*. Nothing tells the assessor to check whether Ingress objects reference the broken controller. A customer mid-outage reads "Impact 1 (tech debt) — fix or remove it." Add a gate: broken **with bound routes** ⇒ flag as suspected active outage (urgent, outside migration scoring); broken with zero bound routes ⇒ tech debt 1.

**S2 — "Cleaned up and deleted" without "export first."** The Migration Crew Alert recommends deleting orphaned Ingress objects — but those manifests are often the only surviving record of routing intent, i.e. the exact input this skill migrates. The skill already exports `current/` YAML (Export Materials). The recommendation should say: export/back up first (link Export Materials), confirm they're not awaiting re-adoption, then delete.

**S3 — "Your cluster is healthy from an ingress standpoint"** (empty-estate note): absence isn't health — an estate that was accidentally torn down also reads "healthy." Reword to state absence and ask the user to confirm it's expected.

**S4 — Empty-estate short-circuit vs mandatory machinery.** §1.0-A says "stop the deduction math," but §1.6 declares the Score Breakdown table MANDATORY and the report template says "follow EXACTLY." State explicitly that the short-circuit replaces §1.1–1.6 for case A. Case A also lacks case B's "no other live findings" guard — a no-ingress cluster can still surface DNS/cert findings from sections 4–5; say which wins.

**S5 — Healthy controller + zero Ingress resources is unhandled — and scores worse than broken.** §1.0-A requires all three absent, so a cluster with only a healthy AWS LBC installed (nothing to migrate) never reaches 100: §1.1 deducts 1–2 for the healthy controller — while a *broken* controller deducts 1. The migration *target* controller, healthy, should not be a deduction at all (consistent with "clean routes count at 0 effort").

**S6 — The scale is now 0–5 but "Impact 1–5" survives everywhere.** SKILL.md Rating Rubric first line, all six shared rating-model headers, §1.1 ("carries an Impact 1–5"), and rendering rule 14 ("🟡1-2 / 🟠3-4 / 🔴5") never admit 0, while §1.0/§1.6 mandate 🟢 0 rows. Pick one framing ("Impact 0–5") and apply it consistently.

**S7 — `base_points` pseudo-code KeyErrors on 0 and Unknown.** `{5: 10, 4: 6, 3: 4, 2: 2, 1: 1}[impact]` in an algorithm labeled "follow EXACTLY"; both 0 and Unknown are assigned 0 pts only in comments. Add them to the dict (`0: 0`) or handle explicitly.

**S8 — "Score the highest-priority dimension" vs "three dimensions stack."** The Impact Indicator says score the highest-priority dimension that applies (pick one); ingress-discovery says the three dimensions "stack… they add up — do not let one override another." These describe different operations on the same finding (one row vs three rows in the Score Breakdown). Clarify: priority order applies *within* a single finding's rating; stacking applies *across* the three finding types on the same object.

**S9 — Business-outranks-security can under-rate pure security findings.** Under strict priority order, an RCE-vulnerable controller serving only an internal admin panel is anchored to its (low) business traffic first — the old "score the dominant dimension" model let security alone reach High. Combined with M1, consider: security findings anchor on exposure/blast-radius, business findings on live traffic, and the priority order resolves *ties*, not *presence*.

---

## ADVISORY (non-blocking)

- **A1** — Empty/orphaned estates reuse the TRIVIAL band, defined as "Mechanical — ALB Controller / ATX auto-converts; hours." Consider a distinct "N/A — nothing to migrate" label so the exec headline doesn't read "migration nearly done."
- **A2** — The bridge/road/"Migration Crew"/"blueprints" metaphors are embedded verbatim in a customer-facing report that is otherwise clinical; consider keeping metaphors in assessor-facing guidance and emitting plain language.
- **A3** — `references/gateway-api.md` still justifies a Low rating by effort ("easy to deploy") — same skill, same rating model, untouched by this PR. The retained header example "an easy-to-deploy prerequisite stays 🟡 low" leans the same way; consider rewording the *reason* (no live-traffic impact) while keeping the outcome.
- **A4** — SKILL.md rubric table puts "present-but-broken controller (tech debt)" in the 1–2 band without pinning "(1)" — every other occurrence pins 1.
- **A5** — SKILL.md instructs "surface [unhealthy controller] as the first finding," but rule 11 orders findings highest-impact-first and broken is now Impact 1 (last). Reconcile.
- **A6** — Two controllers where one is broken: does the broken one count toward the "multiple controllers" Medium? Unspecified.
- **A7** — Pre-existing (not this PR): `misc/evals/eks-ingress-migration/README.md` has a duplicated "Live-MCP caveat" heading with a leftover `<REPLACE:` template placeholder.

---

## Strengths

- Fully resolves #126: a fresh cluster no longer out-scores a healthy estate; traced the arithmetic both ways.
- The deliberate deviation (CrashLoopBackOff 5 → 1) is documented **in the changed files**, not just the PR body — future maintainers will see the reasoning.
- Docusaurus mirror parity is exact across all 9 file pairs; re-running both generators produces zero drift; docs-sync CI green.
- Single clean commit, no unrelated files, no mode-bit changes.
- The HTML renderer already handles 🟢 (badge mapping present) — no renderer work needed.
- The presence-vs-absence distinction, the Impact 0 transparency rows in the Score Breakdown, and the effort-is-not-severity principle are all genuine improvements in the model's direction; the requested changes are about making the edges of that model safe and self-consistent.

kritktongrakroj:
Thanks for the thorough 9-lens review, @utkarpun — it caught real problems. Pushed two commits (e472e43, 5aabd8d)
  addressing everything. Summary below.
  
  Core reframe (fixes M1/M2/M4 together): the root cause was defining the 🟢 0 band by the predicate "serves no live
  traffic." That's now replaced by presence = estate state (absent / empty / orphaned), with severity anchored by priority
  order. Verified all four factual must-fixes against primary sources before writing them.
  
  Must-fixes
  
  - M1 — Non-event scoped to data-plane exposure; added an explicit control-plane carve-out: a reachable known-CVE/EOL
  controller is a security finding even at zero routes. Cited CVE-2025-1974 (CVSS 9.8, admission-webhook RCE, exploitable
  with no Ingress; fixed v1.11.5/v1.12.1). (k8s blog 2025/03/24.)
  - M2 — "No traffic" must be verified, not inferred from pod status: all replicas down for nginx; ALB/target-group state for
  the LBC (control-plane only — the ALB serves while the pod is down/uninstalled). Conservative default: if unverified, treat
  as live.
  - M3 — Orphaned config now routes §1.2 IngressClass findings through the §1.0 short-circuit (0); no more contradictory
  Medium.
  - M4 — Non-event defined by estate state, not the traffic predicate; broken controller's routes are explicitly migratable
  config (scored normally).
  - M5 — Migration Crew Alert parameterized ({N}, {CONTROLLER_CLASS}, {SCORE}/{LABEL}); "verbatim 100/TRIVIAL" restricted to
  the empty-estate case; added export/back-up-before-delete.
  - M6 — Eval expectation → "Impact 0–5"; added a strong #126 regression eval (empty ⇒ 100/TRIVIAL; orphaned ⇒ alert; healthy
  zero-route CVE controller ⇒ still a security finding).
  - M7 — Dropped "plus the effort to change" from the Assessment Summary line.
  - M8 — Corrected both IngressClass facts: classless Ingresses are ignored (silent non-reconciliation; nginx needs
  --watch-ingress-without-class=true, default false, or a default class) — no "implicit default"; multiple defaults ⇒
  deterministic newest-default admission since v1.25 (DefaultIngressClass plugin sorts creationTimestamp desc, lowest-name
  tiebreak; verified in admission.go + PR kubernetes/kubernetes#110974), not "non-deterministic." Pre-1.25 rejected.
  
  One deliberate deviation (S1): I split the broken controller instead of a flat 1 — broken + bound routes = suspected active
  outage (urgent, flagged outside the 0–100 score, since migration difficulty is the wrong lens for a live outage); broken +
  zero routes = tech-debt 1.
  
  Shoulds/advisories: S2 (export-first), S3 (absence ≠ health), S4 (short-circuit replaces §1.1–1.6; DNS/cert-with-no-ingress
  = 0), S5 (healthy target controller = 0), S6 (Impact 0–5 everywhere incl. the six sibling headers), S7 (base_points handles
  0/Unknown), S8/S9 (priority-within-finding vs stacking-across; security anchors on exposure), A1–A7 (incl. A7 eval-README
  <REPLACE:> cleanup). Docusaurus mirror regenerated; docs-sync green.
  
  Two follow-ups tracked separately (do not block this PR):
  
  - #127 — the Tier-B cap vs alb-migration contradiction. Note: #127's suggested wording ("Impact 4 when it requires code
  changes") is effort-based, which conflicts with this PR's "effort ≠ severity" principle — so I'll resolve #127 by
  re-anchoring on faithful-equivalent/business-impact rather than effort. For now I only removed the self-contradiction our
  header created in alb-migration.md.
  - migration-plan.md §7.3 — still rates Impact by timeline and references an "effort dimension" the new model abolished;
  will fold into the #127 cleanup.
  
  Ready for another look when you have a moment — thanks again! 🙏

utkarpun (dismissed):
## Round 2 Review — head 5aabd8d

**Verdict: CHANGES REQUESTED** — substantial, good-faith fix round. 6/8 original must-fixes resolved cleanly; M8 fully confirmed against upstream code (admission.go, flags.go, KEP). Mechanics are merge-ready (zero-drift regen, no conflicts, CI green). But the new "presence = estate state" model has internal contradictions — the two headline fixes (M1 CVE carve-out, M2 verify-don't-infer) are each undone by the §1.0 short-circuit they sit next to.

---

### MUST-FIX (6 items — all localized wording/consistency repairs, no model rework needed)

**R2-M1 — Case A short-circuit swallows the CVE carve-out** (report-generation.md:16-17 vs ingress-discovery.md:104)
Case A covers "the only controller present is a healthy migration-target controller" → "Stop the deduction math. This short-circuit replaces §1.1–1.6." But "healthy" ≠ "not vulnerable": a running CVE/EOL controller at zero routes triggers Case A → 100/TRIVIAL and §1.4 never runs — the exact scenario M1 was filed to catch (eval 3 Cluster C grades on it). **Fix:** Gate Case A with "and no security finding on the controller (not CVE/EOL-affected)".

**R2-M2 — Case B categorically asserts "zero live traffic" with no verification gate** (report-generation.md:23,31)
§1.0-B declares orphaned Ingress objects "carry zero live traffic … no downtime risk" and the Crew Alert hardcodes "handling zero live production traffic" — but ingress-discovery.md:12 (same PR) says the ALB keeps serving while the controller is down or uninstalled. Orphaned alb-class config can front live traffic via the surviving ALB, and the Alert then recommends cleanup. This is the original M2 scenario surviving inside Case B. **Fix:** Route Case B through the "verify before you downgrade" rule (check surviving ALBs/target groups for alb-class orphans; if unverified, treat as live).

**R2-M3 — Scale/Volume formula computes 12, worked example and eval 1 say 10** (report-generation.md:63 vs :161 and evals.json:13)
"Routes that need work = total − 0-effort routes" → 18−6=12 (the 2 annotation-only moves don't meet the 0-effort definition at :52). A compliant model computes 12 and fails eval 1. **Fix:** Align the definition, the example cell, and the eval to one number.

**R2-M4 — DNS/cert non-event precedence exists only in Case A; Case B contradicts eval 3** (report-generation.md:19; dns-certificates.md:28; evals.json:36)
Eval 3's Cluster B (orphaned, nothing else) requires 100/TRIVIAL, but Case B lacks the "DNS/cert items listed at 0" precedence, so "no external-dns" fires 🔴5 → −10 → 90. **Fix:** Hoist the precedence to cover both cases.

**R2-M5 — §7.3 effort residue is not quarantined; feeds the score** (migration-plan.md:147,159-161 vs report-generation.md:37)
"Compile ALL findings from sections 1–7. Every item must appear" pulls §7.3's timeline-scored Impact into the master list. Deferring the §7.3 rewrite to #127 is fine, but add an explicit quarantine line: "§7.1–7.3 outputs never enter the Score Breakdown."

**R2-M6 — High-band CVE rule + eval 3 depend on webhook state discovery never collects** (ingress-discovery.md:108 vs :91-94)
The 🔴5 band requires "with the validating admission webhook enabled", but §1.4 "How to check" only reads image tag + ConfigMap keys — no `kubectl get validatingwebhookconfigurations` / controller-args step. **Fix:** Add the discovery step.

---

### SHOULD-FIX (10 items)

- **R2-S1** — Spec misquote (ingress-discovery.md:62): text says "should ignore"; actual KEP-1453 wording is "may choose to ignore". Soften to "are permitted to ignore".
- **R2-S2** — ALB CORS overstatement (alb-migration.md): ALB response-header modification can inject static CORS headers (Nov 2024). Dynamic origin reflection + preflight short-circuit are what need app changes. Qualify.
- **R2-S3** — WAF rate-window wording: "per-IP over a multi-minute window" → aggregation default per-IP with custom keys; window 60–600s.
- **R2-S4** — M7 residue: effort framing survives in Assessment Summary template (report-generation.md:283,285,291): column header "Why — feature transfer / replacement effort", row template "+ replacement effort", :291 "how hard to replace" — ban says otherwise. Also :354/:361 mandate "Effort: Low/Medium/High" vs :497's ban.
- **R2-S5** — Worked example scores class-switch moves at 0 (report-generation.md:158 vs alb-migration.md:169 "a bare class switch is at least Medium"). Interacts with R2-M3.
- **R2-S6** — Unrouted case: orphans of class X while a DIFFERENT controller is installed. Case B requires "no controller installed"; nginx orphans + healthy ALB controller matches neither case. Route explicitly.
- **R2-S7** — "bound routes" used ~15× but never defined. Define once.
- **R2-S8** — No how-to for the ALB/target-group verification check (load-bearing for R2-M2): every other check has commands; this one has none. Add `aws elbv2 describe-target-health` etc.
- **R2-S9** — Stacking preamble still describes dimension (b) as "tech-debt deduction (present-but-broken, ignored)" without the S1 split (ingress-discovery.md:166).
- **R2-S10** — §1.6 sample table mixes incompatible states: "broken nginx pod, zero bound routes" alongside "12 orphaned Ingress objects" — if same class, those orphans are bound → outage. State the classes differ.

---

### ADVISORY (9 items)

A–I: Case A label ambiguity (two labels, unclear which goes in SCORE tag); Case A extension self-violates precondition (healthy LBC always ships IngressClass vs "no IngressClass" requirement); degraded-but-serving controller has no classification row; unverifiable-traffic bucket unclear; template hygiene (duplicate headings, duplicate Phase blocks, column mismatches); "sections 4–5" never mapped to files; migration-plan.md hardcodes Week 1–5 vs §7.3's weeks-to-months; CVE phrasing alignment opportunity; A7 REPLACE block removal verified closed.

---

### Confirmed

- All hard external facts verified against upstream code: CVE-2025-1974 (9.8, v1.11.5/v1.12.1, 2025-03-24, webhook default-on), --watch-ingress-without-class=false (flags.go), DefaultIngressClass newest-timestamp + lowest-name tiebreak (admission.go v1.33 + PR #110974), ALB-keeps-serving (LBC #1629).
- Contributor correctly followed code over the stale kubernetes.io docs page.
- Score machinery: bands partition cleanly, caps cannot exceed 100 or go negative, Impact 0 breaks no multiplier.

utkarpun:
Heads-up: main has moved since this branch was cut — #134 (ecs-build skill) and #135 (strict frontmatter validation + YAML-correct doc generators) are now merged. Please rebase onto current main when you push your round-2 fixes. Note #135 adds a frontmatter validation gate to docs-sync CI (strict YAML parse, description ≤1024 chars, manifest sync), so run `python3 misc/validate-frontmatter.py` and `./misc/update-all-references.sh --check` + `./misc/update-pages.sh --check` locally after rebasing to catch anything before CI does.

kritktongrakroj:
Thanks again for the detailed round-2 review, @utkarpun. Rebased onto current main (picked up #134/#135) and pushed the fixes — python3 misc/validate-frontmatter.py, ./misc/update-pages.sh --check,
  and ./misc/update-all-references.sh --check all pass locally. Summary below.
  
  Must-fixes
  
  - R2-M1 — Case A short-circuit now gated with "and not CVE/EOL-affected". A running known-CVE/EOL controller at zero routes no longer hits the 100/TRIVIAL short-circuit; it runs §1.4 and scores the
  security dimension + raises the gate (control-plane exposure survives zero routes).
  - R2-M2 — Case B routed through verify-before-downgrade: for alb-class orphans, check surviving ALB/target-group state before calling them dead; unverified ⇒ treat as live. The Migration Crew Alert
  no longer asserts "zero live traffic" unconditionally.
  - R2-M3 — Scale/Volume reconciled to 12 (total − 0-effort = 18 − 6) across the §1.3 definition, the worked example, and eval 1. The 2 annotation-only class-switch moves now count as work.
  - R2-M4 — DNS/cert non-event precedence hoisted into Case B, so an orphaned estate with "no external-dns" still lands 100/TRIVIAL (eval 3 Cluster B).
  - R2-M5 — Added an explicit quarantine (report-generation §1.1 + migration-plan "Checks to Execute"): §7.1–7.3 are planning outputs and never enter the Score Breakdown; §7.3 timeline is no longer
  rated as an Impact.
  - R2-M6 — Added the kubectl get validatingwebhookconfigurations (+ controller-args / admission Service) discovery step to §1.4 that the control-plane CVE band depends on.
  
  Should-fixes
  
  - R2-S1 softened to KEP-1453 "may choose to ignore" · R2-S2 ALB can inject static CORS headers (Nov 2024); only dynamic origin reflection + preflight need app changes · R2-S3 WAF rate rules corrected
  (default per-IP aggregation, custom keys, 60–600s window) · R2-S4 effort framing dropped from the Assessment Summary template; option-panel "Effort" clarified as path-level, not a finding Impact ·
  R2-S5 class-switch moves no longer scored at 0 · R2-S6 Case B made per-class · R2-S7 "bound routes" defined at first use · R2-S8 added the ALB/target-group verification commands · R2-S9 stacking
  preamble split for broken-with-bound-routes · R2-S10 §1.6 sample now uses distinct controller classes.
  
  One deliberate deviation from the review — R2-M8 / §1.2 (multiple default IngressClasses). The previous round accepted "pick newest (code over stale docs)". On re-checking the primary source, the
  current official Ingress concept doc (v1.36, updated 2025-11-24, Default IngressClass (https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class)) still states verbatim:
  "the admission controller prevents creating new Ingress objects that don't have an ingressClassName specified." To keep the indicator grounded in the documented, source-of-truth behavior, §1.2 now
  leads with the doc's rejection behavior and keeps the code nuance (kubernetes/kubernetes#110974) as a version caveat. Happy to adjust if you'd prefer to lead with the code behavior instead.
  
  Ready for another look — thanks!

utkarpun (dismissed):
## Round-3 review — NO-GO (2 must-fixes)

Thanks @kritktongrakroj — strong round-2 pass. Verified all 6 R2 must-fixes in-tree at `a1e12a0`: **M1–M5 landed clean** (both the CVE-2025-1974 zero-route and the ALB-keeps-serving loops are confirmed closed), M6 mostly landed, and 9/10 should-fixes are done. Two things block merge, plus the contested §1.2 ruling goes the other way. None are large.

### Must-fix 1 — §1.2 ruling: lead with pick-newest, not rejection

On the multiple-default-IngressClass deviation you flagged: I ran a live check of `plugin/pkg/admission/network/defaultingressclass/admission.go` across tags v1.25.0 → v1.36.0 and master (as of 2026-07-23). **Every EKS-supported version (1.30–1.36) picks the newest default (CreationTimestamp, name tiebreak) and silently assigns it** — it does not reject. Rejection existed only through v1.24 and was removed by [k8s#110974](https://github.com/kubernetes/kubernetes/pull/110974) (shipped v1.25.0). The kubernetes.io Ingress page (modified 2025-11-24) still describes the pre-1.25 rejection behavior — it's stale.

So `ingress-discovery.md:63` is inverted: it leads with rejection as "source-of-truth" and demotes pick-newest to a caveat. On the versions this skill targets, rejection never happens. **Please flip it**: lead with pick-newest (the silent-assignment / cutover-ambiguity risk), demote rejection to a "≤ v1.24 only" historical note with explicit version numbers, keep the k8s#110974 cite. Your "either way, remediation is exactly one default" framing is good and can stay — just fix the primary/caveat ordering, and stop presenting "blocked Ingress creation" as a live EKS outcome.

### Must-fix 2 — NH-1: the M1 fix contradicts the Re-architecture Gate definition

The new M1 security gate at `report-generation.md:17` orders raising the Re-architecture Gate for a zero-route CVE cluster — but the gate is defined everywhere (`SKILL.md:326`, `report-generation.md:61/:111`) as a count of *routes* needing redesign (`⛔ N routes`). A zero-route cluster has none, so obeying `:17` emits "⛔ 1 route" on a zero-Ingress cluster and then trips the new consistency check at `:189` (gate count ≠ Tier-A findings). The two newly added rules force contradictory output. Fix: either drop "raise the Re-architecture Gate" from `:17` (the 🔴5 security row already carries severity), or define a non-route gate variant.

### Should-fix

- **M6 residual** (`ingress-discovery.md:95/:109`): band keys on webhook *"enabled"* but discovery records *"present/absent"*; no rule when the VWC exists but controller args lack `--validating-webhook`; and **no RBAC-denied fallback** — an assessor who can't read VWCs silently loses the 🔴5 band. Add a safe default.
- **S8** (load-bearing for M2): elbv2 commands are present, but there's no step to find *which* ALB belongs to the orphaned Ingress (`status.loadBalancer.ingress[].hostname` or tag lookup via `ingress.k8s.aws/stack`). Without it the verify-before-downgrade flow isn't executable on a multi-ALB account.
- **NH-2** (`report-generation.md:166`): comment promises a class-switch deduction "counted once below" that no row actually counts — invites an invented deduction and breaks determinism. Drop the dangling comment.
- **eval 3 Cluster B**: fixture never states "no surviving load balancers", so a strict unverified⇒live reading could deny it 100/TRIVIAL. One clause makes it airtight.
- Minor syncs: `dns-certificates.md:28` still says "🔴5: No external-dns" unconditionally; `SKILL.md:323` Case-A summary omits the "not CVE/EOL-affected" qualifier; **NH-3** `:17`/`:36` same-file `§` refs should name `ingress-discovery.md §…`.

### Open question (not blocking)

M1 arithmetic: a zero-route CVE cluster caps at Impact 5 = 10 pts → scores 90, which `§1.5` still bands "90–100 TRIVIAL" — so a CVE cluster wears a TRIVIAL badge alongside the red gate + 🔴5 row. Intended?

Full detail available on request. Nice work on the hard parts — the two original security defects are genuinely closed.

kritktongrakroj:
Thanks for the round-3 pass @utkarpun — both must-fixes and the should-fixes are in at `47f6ebd`.
  
  **Must-fix 1 — §1.2 ruling flipped to lead with pick-newest.** You're right, and this
  **reverses** my round-2 "lead with the doc" compromise. On re-checking the primary source
  your way — the `DefaultIngressClass` admission plugin across tags v1.25 → v1.36 + the merge
  PR kubernetes/kubernetes#110974 — every EKS-supported version (1.30–1.36) **silently assigns
  the newest default** (creationTimestamp, name tiebreak) and does **not** reject. `ingress-discovery.md`
  §1.2 now leads with pick-newest (silent-assignment / cutover-ambiguity risk), demotes rejection
  to a "≤ v1.24 only" historical note with explicit versions, keeps the #110974 cite, and marks
  the kubernetes.io concept page stale for supported versions. "Blocked Ingress creation" is no
  longer presented as a live EKS outcome.
  
  **Must-fix 2 — NH-1 gate contradiction resolved.** The zero-route CVE no longer emits a phantom
  `⛔ N routes`. `report-generation.md` §1.0-A now records it against the Re-architecture Gate's
  **EOL-CVE / control-plane-RCE condition** line (a non-route *condition*); the §1.4 gate CVE line
  is broadened to cover a control-plane RCE at any route count; the Step-2 consistency check now
  reconciles the gate against route findings **plus** non-route conditions (so a security condition
  legitimately raises the gate with zero routes and isn't reconciled away); and the badge reads
  `⛔ N route(s)/condition(s)` in SKILL.md, the template, and the gate label.
  
  **Open question (90/TRIVIAL):** kept the band — the score measures *migration* work and a zero-route
  CVE genuinely has little to migrate — but §1.5 now **requires** the one-line bottom-line to name the
  CVE whenever the gate carries an EOL/CVE condition (e.g. *"90 / TRIVIAL · ⛔ CVE-2025-1974
  control-plane RCE — nothing to migrate, but patch/replace the vulnerable controller urgently"*), so
  "TRIVIAL" is never read alone on a vulnerable cluster. Happy to add a distinct security band instead
  if you'd prefer.
  
  **Should-fixes:**
  - **M6 residual** — §1.4 webhook discovery is now a tri-state (`exposed` / `disabled` / `unverified`)
    with an **RBAC-denied safe default**: can't read the VWC/args ⇒ assume exposed on a `< v1.11.5 /
    < v1.12.1` controller, never silently drop the 🔴5 band. 🔴5 wording aligned to "exposed".
  - **S8** — added the ALB-identification step (`status.loadBalancer.ingress[].hostname`, or tag lookup
    via `ingress.k8s.aws/stack` + `describe-tags`) so verify-before-downgrade is executable on a
    multi-ALB account.
  - **NH-2** — dropped the dangling "class-switch counted once below" comment in §1.7.
  - **eval 3 Cluster B** — fixture now states "no surviving load balancers" so verify-before-downgrade
    confirms it dead ⇒ 100/TRIVIAL.
  - Minor syncs: `dns-certificates.md` 🔴5 "No external-dns" is now a non-event (0) on a dead/absent
    estate; `SKILL.md` Case-A adds the "not CVE/EOL-affected" qualifier; NH-3 same-file §refs at :17/:36
    now name `ingress-discovery.md §…`.
  
  CI green locally: `validate-frontmatter.py` (29 skills, 0 errors), `update-pages.sh --check`,
  `update-all-references.sh --check`; Docusaurus mirror regenerated. Ready for another look — thanks!

utkarpun:
Round-4 delta verified and **merged** — thanks @kritktongrakroj. 🎉

Both round-3 must-fixes landed cleanly:
- **§1.2** now leads with pick-newest (silent newest-default assignment on all EKS-supported versions, v1.25+), rejection correctly demoted to a ≤ v1.24 historical note with the k8s#110974 cite.
- **Re-architecture Gate** — the contradiction is closed: the gate algorithm, consistency check, and badge all reconcile route findings plus non-route conditions, and the scenario traces (zero-route CVE, normal multi-route, orphaned-only) all pass.

All four original defect loops (CVE-2025-1974 zero-route, ALB-keeps-serving, DNS/cert non-event, §7 quarantine) are substantively closed, and every should-fix landed.

Two small residuals moved to #127 as follow-ups (both non-blocking, fail toward over-caution) — they land naturally with your next PR in this area:
1. `report_to_html.py` `_gate_badge` still prints route-only wording, so a zero-route CVE cluster's rendered HTML reads "⛔ 1 route need redesign" (+ an n=1 grammar bug); plus two stale route-only prose sites (`report-generation.md:142`/`:251`).
2. The M6 webhook tri-state keys on `--validating-webhook=false`, which isn't a real ingress-nginx flag form (it's a string address flag), and needs a VWC-entirely-absent branch.

Details + nits are in #127. Appreciate the fast, thorough turnaround across four rounds.

PR #135, opened 2026-07-11
ci: strict frontmatter validation + YAML-correct doc generators (fixes #131)

### LINKED ISSUE #131: CI gap: frontmatter is never strict-parsed — regen corruption ships green
## Problem

The regen scripts (`misc/update-pages.sh`, `misc/update-skills-references.sh`) extract SKILL.md frontmatter with a line-based regex, not a YAML parser. CI's docs-sync check only verifies committed files match regeneration output — so when the regex mis-extracts, the corrupted value is regenerated consistently and CI passes green.

This shipped real corruption three times during the ECS suite onboarding:

1. **PR #116** — a `>-` block-scalar description was extracted as the literal string `">-"` into `skills.json`, the website wrapper, and `skills/README.md`. The live site showed `>-` as the skill description.
2. **PR #122** — escaped inner quotes (`\"`) in a double-quoted scalar survived into `skills.json` as 18 literal backslashes, and pushed the rendered description to 1037 chars (over the 1024 Agent Skills cap) while the parsed value was compliant.
3. **PR #130 assembly** — the same escaped-quote pattern resurfaced on ecs-architect after a review fix and was caught only by manual validation.

Separately, nothing enforces the 1024-char description cap: 3 of the 4 original ECS PRs (#116 #117 #119) exceeded it at first submission.

## Proposal

Add a CI job (or extend docs-sync) that, for every `skills/*/SKILL.md`:
- strict-parses the frontmatter with `yaml.safe_load` (fail on parse error);
- asserts `len(description) <= 1024`;
- asserts the parsed description is byte-identical to the corresponding `skills.json` manifest entry (catches extraction drift regardless of quoting style).

~15 lines of Python; would have caught all three incidents and the length violations.

utkarpun:
## What

Closes #131 across **all three surfaces** the incidents shipped on (skills.json, Docusaurus wrappers, skills/README.md) — not just the manifest.

### Commit 1 — CI validator (`bb5de7a`)
- `misc/validate-frontmatter.py`: strict-parses every `skills/*/SKILL.md` + `devops-agent/*/SKILL.md` frontmatter with `yaml.safe_load` — valid mapping, `name` + `description` required (string), description ≤ 1024 chars, and (for `skills/`) byte-match against the generated `skills.json`; per-entry manifest length/type checks; per-file error isolation (one bad file never hides another)
- `build_manifest` in `update-pages.sh` switched from naive regex to `yaml.safe_load` — the manifest is now correct-by-construction; validator is the drift guard
- Wired as the first step of docs-sync (validate → regen `--check`)

### Commit 2 — pipeline-wide expansion (`085e9f1`)
- `parse_frontmatter()` in **all 5** `misc/update-*.sh` generators now YAML-parses (was naive awk that truncated folded scalars at line 1 and leaked quoting artifacts) — this is what protects the wrapper + README surfaces
- Examples manifest parser converted; both manifest heredocs fail with a clean one-line error on malformed input; pyyaml preflight guard in every generator
- Validator hardening: per-scope duplicate-name detection, empty-description warning, dedicated unreadable-file handler
- `docs-sync.yml`: self-trigger path, pinned `pyyaml>=6,<7`, split validator/regen failure summaries, `head_ref` via env, absolute CONTRIBUTING link
- `vendor-update.yml`: setup-python + pinned pyyaml, removed `|| true` so regen failures fail the job instead of shipping stale docs
- Docs: CONTRIBUTING.md + misc/README.md cover the gate and the local pyyaml prerequisite
- Regen fallout (intentional): 4 devops-agent wrapper pages now show their **full** descriptions (old awk truncated at line 1 — live bug on the site today); workflow-skeleton wrapper description is empty by design (its `---` block is template body, not frontmatter)

## Verification

Four adversarial review rounds + empirical incident replay:
- All 3 historical incidents (`>-` block scalar, escaped-quote 1037/1006 divergence, escaped-quote corruption) replayed end-to-end — now caught or rendered correctly on **all three surfaces**
- 27/27 skills validate, zero-diff idempotent regen, Docusaurus production build passes
- Old-vs-new parser differential-tested over every real input file: only intentional diffs
- Workflow failure paths traced (validator-fail vs regen-fail summaries are mutually exclusive); fork-PR safe (`pull_request`, `contents: read`, no secrets)

## Notes for reviewers
- `ecs-build`'s description parses to 1017 chars — 7 under the cap; the next edit to it may need trimming
- Full regen wall clock is ~76s (~64s update-pages.sh) due to per-file python spawns — acceptable in CI, noted for awareness

### Commit 3 (`95b16df`) — fail-fast + parser convergence
- Invalid YAML in a present frontmatter block now aborts every generator with the file named (exit 2, `inherit_errexit`) — no more silent empty-description wrappers or stray `.tmp` on broken input
- `run_triggering.py` (evals) now uses the canonical YAML parse (regex fallback kept for no-pyyaml envs); old parser leaked surrounding quotes into 6 descriptions
- steering-workflow-creator `quick_validate.py`: new W006 warning when frontmatter isn't at line 1 (the docs pipeline would silently skip such a file)
- Remaining known divergence: `skills/skill-creator` (vendored, upstream-owned) — upstream-sync material only

### Commits 4–6 — post-review hardening (`5182f32`, `7091eaf`, `c0a14e7`)
- `5182f32`: W006 also catches an indented opening `---`; regex-fallback quote-leak divergence documented in the `run_triggering.py` docstring
- `7091eaf`: dedupe eks-cost-intelligence Docusaurus route/redirect (kills the duplicate-route build warning); SIGPIPE guards on dry-run previews
- `c0a14e7`: README slug hook anchored to the docs root via `path.relative` — old regex matched the **first** `/docs/` in the absolute checkout path, producing wrong slugs when the checkout path contains a `docs` segment (hostile-path build-tested)

### Commit 7 — merge-readiness panel fixes (`bb4fc70`)
6-lens adversarial panel findings, folded in rather than deferred:
- **Unclosed frontmatter fails fast** in all 6 parsers (was: silently dropped from generated output with green CI — a regression the strict parsers introduced vs the old lenient ones); unreadable files that open frontmatter also error
- **Validator rejects duplicate frontmatter keys** via a strict SafeLoader subclass (`yaml.safe_load` silently last-wins; spec-conformant js-yaml ≥4 consumers reject — exactly the malformed-input class this gate exists for)
- `vendor-update.yml`: capture-first `ls-remote` + empty-SHA guard (transient failure could overwrite `.vendor-sha` with an empty string, corrupting change detection); regen scripts invoked unconditionally via `bash` (the `[ -x ]` guard skip-greened on a lost exec bit)
- dry-run previews capture-first instead of `| head || true` (the `|| true` masked a broken devops-agent/examples skill's exit 2 in preview mode; CI `--check` was unaffected)

Panel residuals accepted (documented, non-blocking): local regen on broken input can leave a stray `.tmp` from an earlier pipeline stage (CI unaffected); StrictLoader also rejects YAML merge keys (`<<:`) that generators accept — validator-stricter, no repo file uses them; validator globs one level (`skills/*/SKILL.md`) — flat layout only, matches current repo; fork-PR fix-link points at base repo (404 for fork branches, guidance-only).

### DIFF
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -27,6 +27,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+      - name: Validate SKILL.md frontmatter
+        run: pip install pyyaml && python3 misc/validate-frontmatter.py
       - name: Regenerate and fail on diff
         run: |
           ./misc/update-all-references.sh --check
@@ -44,6 +49,8 @@ jobs:
           ./misc/update-all-references.sh && ./misc/update-pages.sh && git add -A
           git commit -m "docs: regenerate reference tables and pages"
           ```
+
+          If the failure is from `validate-frontmatter.py`, check your `SKILL.md` frontmatter YAML syntax and description length (max 1024 chars).
 
           Full details: [`CONTRIBUTING.md` → Keeping docs in sync](../blob/${{ github.head_ref || github.ref_name }}/CONTRIBUTING.md#keeping-docs-in-sync-auto-generated-catalogues).
           EOF
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -10,6 +10,7 @@ on:
       - "devops-agent/**"
       - "README.md"
       - "CONTRIBUTING.md"
+      - ".github/workflows/docs-sync.yml"
   push:
     branches:
       - main
@@ -31,13 +32,35 @@ jobs:
         with:
           python-version: '3.x'
       - name: Validate SKILL.md frontmatter
-        run: pip install pyyaml && python3 misc/validate-frontmatter.py
+        id: validate
+        run: pip install 'pyyaml>=6,<7' && python3 misc/validate-frontmatter.py
       - name: Regenerate and fail on diff
+        id: regen
         run: |
           ./misc/update-all-references.sh --check
           ./misc/update-pages.sh --check
-      - name: How to fix (shown on failure)
-        if: failure()
+      - name: How to fix frontmatter validation (shown on failure)
+        if: failure() && steps.validate.outcome == 'failure'
+        run: |
+          cat >> "$GITHUB_STEP_SUMMARY" <<'EOF'
+          ## SKILL.md frontmatter failed validation
+
+          `misc/validate-frontmatter.py` strict-parses the frontmatter of every `skills/*/SKILL.md` and `devops-agent/*/SKILL.md`. Check the job log above for the exact file and error, then verify:
+
+          - the frontmatter is valid YAML (a mapping between `---` delimiters — watch unquoted `:` and stray quotes)
+          - `name` and `description` keys are present
+          - `description` is at most 1024 characters
+          - for `skills/*/SKILL.md` only: `description` matches the generated `skills.json` manifest (`devops-agent/` is exempt; a missing manifest entry is only a warning) — if you changed a description, regenerate:
+
+          ```bash
+          ./misc/update-all-references.sh && ./misc/update-pages.sh && git add -A
+          git commit -m "docs: regenerate reference tables and pages"
+          ```
+          EOF
+      - name: How to fix stale docs (shown on failure)
+        if: failure() && steps.regen.outcome == 'failure'
+        env:
+          BRANCH_REF: ${{ github.head_ref || github.ref_name }}
         run: |
           cat >> "$GITHUB_STEP_SUMMARY" <<'EOF'
           ## Docs auto-gen blocks are stale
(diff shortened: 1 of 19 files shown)


utkarpun:
**Final blast-radius review — known non-blocking findings** (documented for the record; verified none block merge):

1. **Local regen is non-atomic on broken frontmatter.** If a SKILL.md has invalid YAML, `update-pages.sh` hard-fails at the manifest builder (clean one-line error) — but the per-wrapper pass runs earlier and silently writes that skill's wrapper with `description: ""`, leaving a partial tree + stray `skills.json.tmp`. CI is unaffected (job fails, nothing commits); local users should `git checkout` after fixing the YAML.
2. **Full regen wall-clock is ~76s** (per-file python subprocess spawns), not the ~30s estimated in the PR body. Fine for CI; worth knowing if regen ever runs in a loop.
3. **Three non-canonical frontmatter parsers remain** (intentionally out of scope — none feed published output): `misc/evals/scripts/run_triggering.py` (own regex parser), `skills/skill-creator/scripts/quick_validate.py` (stricter than canonical), `skills/steering-workflow-creator/scripts/quick_validate.py` E001 (tolerates a leading HTML comment that the canonical line-1 contract skips — a future workflow copied from the skeleton with the comment intact would pass that linter but be skipped by `update-steering-references.sh`).

Also for awareness: `ecs-build`'s description is 1017/1024 chars — 7 under the new cap; the next edit to it may need trimming.

utkarpun:
Update: findings 1–3 from the comment above are now **fixed in commit 95b16df** (fail-fast generators, evals parser convergence, W006 lint warning) and the timing note in the PR body is corrected to ~76s. The only remaining divergence is `skills/skill-creator`'s vendored validator — upstream-owned, excluded by repo vendoring policy. CI green on the new commit.

Issue #2, opened 2026-03-24
Create Troubleshooting Steering Workflow + Example


utkarpun:
Create Troubleshooting Steering Workflow and the corresponding example

utkarpun:
Deprioritized — may revisit later.

PR #118, opened 2026-07-08
feat(skills): add ecs-operation-review skill

jalawala:
## Summary

Adds `ecs-operation-review` — the umbrella Day-2 audit that scores a live ECS estate GREEN/AMBER/RED across 8 best-practices domains and produces a rated report with prioritized actions, deferring to the deep skills for depth. SKILL.md + 10 references + scoring rubric + report_to_html.py.

Part of the ECS skill suite mirroring the EKS skills (GTM Containers SA effort). Authored via the `/apex:new-skill` greenfield flow; every factual claim was live-verified against public AWS docs (ECS Best Practices Guide + Developer Guide) on 2026-07-08 and enriched via the Containers specialist knowledge base. Unverifiable claims were softened/deferred rather than asserted.

## If this PR adds or changes a skill

- [x] Walked the equivalent manual steps in `CONTRIBUTING.md` (greenfield authoring + GATE-1 currency + 8-lens self-review)
- [x] `skills/<skill>/SKILL.md` present with valid frontmatter
- [ ] `misc/evals/<skill>/triggering.json` — **deferred to a follow-up eval PR** (evals are not a CI gate; batching the eval set + bidirectional SIBLING_MAP once the full ECS suite lands)
- [ ] Neighbour `SIBLING_MAP` updates — **deferred to the same follow-up** (several ECS siblings, incl. `ecs-recon`, are not yet in-repo; sibling-map fan-out will run when they land)
- [x] Regenerated docs (README + skills/README Skills Reference marker blocks, Docusaurus wrappers/manifests) and committed them — `docs-sync` CI is green

## Notes for reviewers

- **Scope:** self-contained content PR (skill + regenerated catalogs/pages). No `steering/ecs.md` hub or `/apex:ecs-*` command shim yet — those come with the suite-level wiring PR.
- **`ecs-recon` routing:** references to the not-yet-merged `ecs-recon` skill are softened ("once available") so nothing dangles.
- **Docs regen on macOS:** used GNU awk (`gawk`) to match CI's Ubuntu awk; both `update-all-references.sh --check` and `update-pages.sh --check` pass locally.

Issue #133, opened 2026-07-10
EKS skill descriptions lack ECS exclusions — routing non-determinism with the new ecs-* name-twins


utkarpun:
## Problem

The new ecs-* skills all carry explicit "Skip for EKS/Kubernetes (use eks-*)" exclusions in their descriptions, but the pre-existing EKS-side descriptions were written before the ECS family existed and mostly lack the reverse exclusion. Ambiguous prompts can now route to either family non-deterministically.

Observed in trigger simulations during the ECS PR reviews:
- "review my cluster operations" — coin-flips between eks-operation-review and ecs-operation-review (worst case: name-twins with bidirectional ambiguity)
- "Fargate or EC2 for my containers?" — can land on eks-design or ecs-architect
- GPU/LLM prompts ("which launch type for GPU inference") — eks-genai has no non-EKS exclusion while claiming broad GPU/Neuron decision scope

## Ask

Add ECS exclusions ("Skip for Amazon ECS — use ecs-<twin>") to the descriptions of, in priority order:
1. **eks-operation-review** (highest risk — direct name-twin)
2. **eks-genai** (broad hardware-decision claims, direct twin)
3. **eks-security** (has a partial non-EKS exclusion; verify it names ecs-security)
4. **eks-design / eks-best-practices** (generic container-architecture pull)

Each edit must respect the 1024-char description cap and re-run the regen scripts (see #131 for why validation matters here).

Note: SIBLING_MAP / triggering.json fan-out for these twins is tracked separately under the eval-infrastructure work (#98) — this issue is only the description-level routing fix.

PR #137, opened 2026-07-13
feat(skills): add ecs-recon skill

literalice:
## Summary

Adds `ecs-recon`, a read-only Amazon ECS environment reconnaissance and discovery skill (the ECS analogue of `eks-recon`). It discovers clusters, services, and tasks across 10 modules — overview, compute & capacity providers, task definitions, deployment configuration, auto scaling, networking, security posture, observability, IaC, and CI/CD — and emits a structured `.ecs-recon-report.yaml`. All operations are strictly read-only (Describe/List/Get).

## What's included

- `skills/ecs-recon/SKILL.md` — orchestration, activation guidance, modules table, CLI (and future MCP) command reference, and the report schema
- `skills/ecs-recon/references/*.md` — 10 self-contained detection modules
- Regenerated docs via the repo scripts (`update-all-references.sh` + `update-pages.sh`): README skills tables, `skills/README.md` detail, Docusaurus site pages, and manifests

## Scope boundaries

Discovery only. Defers to siblings: scoring/audit → `ecs-operation-review`; design, launch-type selection & best practices → `ecs-architect`; deployment & CI/CD → `ecs-devops`; GPU/ML → `ecs-genai`; security & compliance → `ecs-security`; cost/TCO → `ecs-cost-intelligence`; observability design → `ecs-observability`; replatform/migration → `ecs-modernize`; EKS → `eks-recon`.

## Testing

- Read-only audit of every CLI/MCP command across all 10 references (corrected the CodeDeploy CLI service name to `aws deploy`)
- Description triggering reviewed against the real ECS sibling scopes
- Docs regenerated with the repo scripts; `docs-sync --check` parity verified on a clean tree

- [x] `skills/ecs-recon/SKILL.md` present with valid frontmatter
- [x] Ran the docs regeneration scripts (`update-all-references.sh` + `update-pages.sh`) and committed the resulting changes (regenerated wrappers/manifests and marker-block updates); `--check` parity verified
- [ ] `misc/evals/ecs-recon/triggering.json` — not included. Follows the current ECS skill precedent (no `misc/evals/` dir, consistent with recently merged ECS skills #118/#119/#123). Happy to add a `triggering.json` eval set if maintainers prefer.


skills/ecs-recon/references/networking.md
```diff
@@ -0,0 +1,527 @@
...
+)
+-> Extract serviceConnectConfiguration
+```
+
+**CLI:**
+```bash
+aws ecs describe-services \
+  --cluster <cluster-name> \
+  --services <service-name> \
+  --query 'services[0].serviceConnectConfiguration'
```
jalawala: **[Must Fix]** `services[0].serviceConnectConfiguration` returns `null` every time — `serviceConnectConfiguration` isn't a member of the `Service` object, so Service Connect is reported absent even when active (a false negative in a discovery skill). It lives on the **Deployment** object; read it from the PRIMARY deployment:

```
services[0].deployments[?status=='PRIMARY'] | [0].serviceConnectConfiguration
```

Same fix applies to the MCP block (~line 278) and the decision-logic bullets at lines 54 and 234. Verified against [API_Service](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Service.html) (no such field) and [API_Deployment](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Deployment.html) (`serviceConnectConfiguration` is a member), and confirmed the corrected JMESPath extracts it from the PRIMARY deployment.

skills/ecs-recon/references/compute.md
```diff
@@ -0,0 +1,382 @@
...
+
+```
+1. Cluster Capacity Providers  -> Enumerate providers associated with the cluster
+2. Service Compute Model       -> Determine launch type vs capacity provider strategy per service
+3. Task Counts                 -> Collect running/desired/pending counts per service
+```
+
+**Why this order matters:**
+- Cluster-level capacity providers establish the available compute pool — services reference these
+- A service either specifies an explicit `launchType` (FARGATE or EC2) **or** a `capacityProviderStrategy` — never both
```
jalawala: **[Should Fix]** `launchType` valid values are `EC2 | FARGATE | EXTERNAL | MANAGED_INSTANCES` — the classification here (and the schema comments at ~line 253 and `overview.md` line 323) only handle FARGATE/EC2, so a Managed Instances or External service gets mislabeled. The family covers Managed Instances and ECS Anywhere/External heavily, so recon should classify them. Source: [API_Service — launchType](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Service.html).

skills/ecs-recon/SKILL.md
```diff
@@ -0,0 +1,421 @@
...
+**Also trigger this skill when:**
+- Another workflow needs ECS environment information as input
+- User mentions ECS and needs discovery before recommendations
+- You need to understand the ECS setup before giving guidance
+
+**Do NOT use this skill for:**
+- **Cost scoring or efficiency analysis** — belongs to `ecs-cost-intelligence`
+- **Security auditing or compliance scoring** — belongs to `ecs-security`
+- **Best-practices evaluation or maturity ratings** — belongs to `ecs-best-practices`
+- **Migration planning (ECS-to-EKS or version migration)** — belongs to `ecs-migration`
```
jalawala: **[Should Fix]** `ecs-migration` doesn't exist (no sibling references it). The family uses `ecs-modernize` for replatform/refactor — which your own description on line 3 already names correctly. Suggest routing here to `ecs-modernize` (hedged "(once available)" per the family convention).

skills/ecs-recon/SKILL.md
```diff
@@ -0,0 +1,421 @@
...
+
+**Also trigger this skill when:**
+- Another workflow needs ECS environment information as input
+- User mentions ECS and needs discovery before recommendations
+- You need to understand the ECS setup before giving guidance
+
+**Do NOT use this skill for:**
+- **Cost scoring or efficiency analysis** — belongs to `ecs-cost-intelligence`
+- **Security auditing or compliance scoring** — belongs to `ecs-security`
+- **Best-practices evaluation or maturity ratings** — belongs to `ecs-best-practices`
```
jalawala: **[Should Fix]** `ecs-best-practices` doesn't exist, and this contradicts your own description on line 3, which routes ECS best practices to `ecs-architect` (it absorbed that corpus). Suggest changing to `ecs-architect`.

skills/ecs-recon/references/security.md
```diff
@@ -0,0 +1,365 @@
...
+    secret_name: CONFIG_TOKEN
+    source: ssm_parameter_store
+    value_from: "/prod/config-token"
+```
+
+### ECS Exec requires both service and task definition enablement
+
+For ECS Exec to be fully functional, three conditions must be met:
+1. The service must have `enableExecuteCommand: true`
+2. The task definition must include the SSM agent sidecar (or use a platform version that includes it)
```
jalawala: **[Should Fix]** ECS Exec itself doesn't require adding an SSM agent sidecar to the task definition — ECS injects and manages the SSM Session Manager binary automatically. The actual task-def requirements are: a task role with SSM Session Manager permissions, a **writable** root filesystem (`readonlyRootFilesystem: true` is unsupported), and `enableExecuteCommand` on the service. Suggest dropping the "must include the SSM agent sidecar" line. (Note: Service Connect and Runtime Monitoring *do* run as sidecars per the same doc — just not ECS Exec.) Source: [ECS Exec](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html).

jalawala (changes requested):
Thanks Hayashi — this is a well-built discovery skill. Read-only discipline is spot on (only describe/list/get), it stays cleanly in the discovery lane without straying into scoring or design, and the pagination handling is thorough. I spot-checked a good sample of the CLI commands and field interpretations against the live AWS API reference (2026-07-13) — Container Insights states, circuit-breaker fields, autoscaling metrics, CodePipeline ECS vs CodeDeployToECS, the `aws deploy` CLI subtlety, launchType/capacityProvider mutual exclusivity — all correct.

One functional bug worth fixing before merge (Service Connect will never be detected — #1), plus a few should-fixes on launch-type coverage and two routing targets that don't exist. Details inline. None of this is a design problem — the bones are solid.

A couple of reasonable v1 deferrals for the backlog (non-blocking): no container-instance drill-down (`describe-container-instances` → `agentConnected`/agent version) for EC2/MI estates, no `runtimePlatform` capture (Graviton detection), and no GPU surfacing (`resourceRequirements` type `GPU`) for the ecs-genai handoff.

utkarpun (changes requested):
Reviewed at head 46c32a3 (2026-07-14). 9-lens pipeline + per-finding adversarial verification; every finding below survived a refutation attempt against live AWS docs (source links inline). JP's round-1 fixes all verified landed correctly — thank you for the fast turnaround.

## 9-Lens Review: CHANGES_REQUESTED

First, the headline: this is a genuinely well-built discovery skill. Read-only discipline is flawless (all 19 distinct AWS calls are List/Describe/Get; secrets reported by ARN/name only, never values), the structural side is exemplary (already rebased onto main, passes the new strict frontmatter CI from #135, docs-regen parity green, mirror parity clean), and the report-shape conventions genuinely mirror eks-recon — some improved on it. The must-fixes below are concentrated in two classes: **silent false negatives** (a discovery skill reporting wrong current-state without erroring — same class as round 1's Service Connect bug) and **stale/incorrect factual claims**.

### INCORRECT (must fix)

| # | File | Claim / defect | Fix | Source |
|---|------|----------------|-----|--------|
| 1 | SKILL.md:40 + all 10 references ("MCP (future)" blocks) | "No ECS MCP server currently exists" — false as of 2026-07-14. The fully managed Amazon ECS MCP server launched in preview Nov 2025; awslabs `ecs-mcp-server` has been public since 2025 (now marked legacy in favor of the managed one). The ~29 speculative tool signatures (`ecs_list_clusters`, `list_ecs_clusters` — two naming styles that already drifted apart internally) match none of the real tools. Mitigation acknowledged: the managed server's 8 tools are diagnostics-shaped (all require `cluster_name`; none enumerate clusters), so your CLI-first design is still right — only the absence claim and invented signatures need fixing. | Replace the "future MCP" sections with a short note: managed ECS MCP server exists (preview) but offers no cluster-enumeration tools, so CLI mode is primary; link the doc page. Drop the invented signatures. | [ecs-mcp-introduction](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-mcp-introduction.html), [tool list](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-mcp-tool-configurations.html), [What's New Nov 2025](https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-eks-ecs-fully-managed-mcp-servers-preview/) |
| 2 | references/deployment.md:67-98, 266-289, schema :246-262 | Services using ECS-native blue/green, linear, or canary (GA 2025-07-17) keep `deploymentController.type == "ECS"` but set `deploymentConfiguration.strategy` — the module classifies on controller type alone, so they're silently misclassified as `ecs_rolling`, dropping `strategy`, `bakeTimeInMinutes`, `lifecycleHooks`, `alarms`, canary/linear config. | Extract `deploymentConfiguration.strategy`; classify ECS+ROLLING → ecs_rolling, ECS+BLUE_GREEN\|LINEAR\|CANARY → native strategy verbatim; add `strategy`/`bake_time_in_minutes`/`alarms` to the schema. | [API_DeploymentConfiguration](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DeploymentConfiguration.html), [What's New Jul 2025](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-ecs-built-in-blue-green-deployments/) |
| 3 | references/deployment.md:155-173, 297-324 | For CODE_DEPLOY/EXTERNAL controllers the module reads only `deployments[]` — but the API docs are explicit that Deployment "is used only when a service uses the `ECS` deployment controller type"; blue/green rollout state for those controllers lives in `Service.taskSets[]` (PRIMARY/ACTIVE, `stabilityStatus`, `scale`), which is never queried. In-flight rollout data available in the same DescribeServices response is silently dropped. | For CODE_DEPLOY/EXTERNAL, additionally extract `services[0].taskSets[*].{id,status,stabilityStatus,taskDefinition,computedDesiredCount,runningCount}`. | [API_Deployment](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Deployment.html), [API_TaskSet](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskSet.html) |
| 4 | references/compute.md:114-117, 264-267 | ECS Managed Instances capacity providers are unclassifiable: the query projects only `{name,status,autoScalingGroupProvider}` and the classification (FARGATE/FARGATE_SPOT by name, else ASG by `autoScalingGroupProvider`) has no rule an MI provider matches — `CapacityProvider.type` is now a first-class field with enum `EC2_AUTOSCALING \| MANAGED_INSTANCES \| FARGATE \| FARGATE_SPOT` and MI providers carry `managedInstancesProvider`. Given the family's MI coverage (and round 1's launchType fix), recon should classify them. | Add `type` and `managedInstancesProvider` to the query; classify on the API's `type` field with an `unrecognized` fallback. | [API_CapacityProvider](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CapacityProvider.html) |
| 5 | references/overview.md:93, 96, 147, 150, 343-352 | Pagination guidance silently truncates estates >100 services: (a) AWS CLI v2 auto-paginates, so the documented "if `nextToken` is present, repeat with `--starting-token`" loop never fires (verified: `list-clusters` emits no token key); (b) when the recommended `--max-items 100` IS used, the CLI emits `NextToken` (capital N — verified live), so an agent looking for lowercase `nextToken` per the doc sees nothing and stops at 100. | Simplest correct fix: drop `--max-items` and rely on CLI v2 auto-pagination; otherwise document the `NextToken` key + `--starting-token` resume flow. | [CLI pagination](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html) |
| 6 | references/security.md:260-261 | "containers cannot call AWS services unless the execution role or instance role (EC2 launch type) provides credentials" — the execution-role half is wrong and security-misleading: execution-role permissions "aren't directly accessible by the containers in the task." The only documented no-task-role path is the EC2 instance role via IMDS. | Drop the execution-role clause; keep the instance-role path. | [Task execution IAM role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html) |
| 7 | references/autoscaling.md:342 | "predictive scaling for ECS is not currently available but may be added" — predictive scaling for ECS service auto scaling has been available since Dec 2024 (dedicated developer-guide pages; `PolicyType` enum includes `PredictiveScaling`). The `unrecognized` fallback keeps behavior safe, but a predictive policy is reported with null metric/target today. | Add a `predictive_scaling` classification (as of 2026-07-14) and drop the absence claim. | [Predictive scaling for ECS](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/predictive-auto-scaling.html) |
| 8 | SKILL.md:8, :100 vs all modules | Scope claim vs reality: SKILL.md promises "account-wide inventory of clusters, services, **tasks**" and "comprehensive context about clusters, services, and tasks" — but no module inventories tasks. Standalone tasks (`RunTask`) and scheduled tasks (EventBridge rules / EventBridge Scheduler ECS targets) are invisible; a batch/cron estate gets a complete-looking report omitting entire workloads. | Either (a) add a standalone/scheduled-task step (ListTasks per cluster, flag tasks whose `group` isn't `service:*`; check EventBridge/Scheduler ECS targets), or (b) scope the claim and report metadata to "services only; standalone/scheduled tasks not discovered." Either resolves it; (a) is the durable answer. | [ListTasks](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListTasks.html), [Scheduled tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/scheduled_tasks.html) |
| 9 | All 11 files | Zero sourcing: no `## Sources` section in any file, zero `docs.aws.amazon.com` links across 4,600 lines, zero dated freshness gates. Family rule (applied to every prior ECS skill): factual claims cite public AWS sources; stale-prone tables carry "Facts verified YYYY-MM-DD" gates. Several findings above (MCP, predictive scaling) are exactly the drift this rule exists to catch. | Add a Sources section per reference (or consolidated in SKILL.md) covering the consequential claims; date-gate the stale-prone enum/feature tables (launchType, Container Insights states, controller table, PolicyType table, batch limits). | family convention, cf. merged ecs-* skills |

### IMPRECISE (should fix)

| # | File | Issue | Source |
|---|------|-------|--------|
| 10 | compute.md:361-374, security.md:335-348, networking.md:497-510 vs task-definitions.md:245-253, deployment.md:356-363, observability.md:363-381, overview.md:369 | Partial-failure handling contradicts itself across modules: three modules mark the entire module `unavailable: true` when one resource's describe fails (discarding every other service's data), three record a per-service error and continue, and overview's own rule says "never discard already-collected inventory." networking.md is self-contradictory within one section. Adopt one normative pattern: per-resource `error` records + module-level `unavailable` only for total failure. | — |
| 11 | task-definitions.md, observability.md, deployment.md schemas + SKILL.md:230-264 | The per-service `error:` field used in edge-case examples is absent from every Output Schema, including the aggregate report schema — schema-validating consumers miss all failure records. Add `error: string \| null`. | — |
| 12 | compute.md:153-156; autoscaling.md:407-413 | Batching shown without the caps: DescribeServices accepts max 10 services per call (>10 = ClientException mid-recon); DescribeScalableTargets max 50 resource IDs. Overview already documents the 100-cluster analogue — mirror that. | [API_DescribeServices](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeServices.html) |
| 13 | overview.md:188-191 | Example statistics include `drainedEC2TasksCount` and `activeFargateTasksCount`, which don't exist (no "drained" task statistic at all). Agents keying parsing off the example find nothing. Use the documented names. | [API_Cluster statistics](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Cluster.html) |
| 14 | deployment.md:240, :350-354 | Two rolloutState misreadings: (a) FAILED occurs only when the service fails to stabilize **and** circuit breaker is on — without the breaker it stays IN_PROGRESS, so "circuit breaker triggered or tasks could not stabilize" implies a nonexistent path; (b) null rolloutState isn't "a deployment from before the feature existed" — documented causes are non-ECS controller or Classic Load Balancer. | [API_Deployment rolloutState](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Deployment.html) |
| 15 | iac.md:210-213, :229-234 | (a) Stack-status filter omits `IMPORT_COMPLETE`/`IMPORT_ROLLBACK_COMPLETE` — imported never-updated stacks are skipped, producing a false `undetermined`. (b) DescribeStackResources returns only the first 100 resources ("use ListStackResources instead" per the doc) — large CDK stacks silently miss ECS resources. A cheaper exact alternative for services: `describe-stack-resources --physical-resource-id <service-arn>` reverse lookup (note: works for service ARNs; cluster physical IDs are names, not ARNs). | [API_DescribeStackResources](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackResources.html) |
| 16 | security.md:279 | "Without executionRoleArn the ECS agent cannot pull images from private ECR or publish logs" — true for Fargate/Managed Instances/external only; on EC2 the container-instance role covers both. Scope the claim by launch type. | [Task execution IAM role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html) |
| 17 | observability.md:103, :305-309 | Enhanced Container Insights described as adding "metrics, logs, and traces" and "integration with CloudWatch Application Signals" — the docs describe task/container-level metric granularity plus log correlation; no traces, and Application Signals is a separate feature requiring its own instrumentation (as of 2026-07-14). | [Enhanced observability metrics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-enhanced-observability-metrics-ECS.html) |
| 18 | overview.md:328 + SKILL.md:287 | `stopped_tasks: int` with "0 if not retrievable" conflates zero with unknown (the prose note shows you're aware — but the emitted data can't disambiguate). Type it `int \| null` like sibling fields. | — |
| 19 | SKILL.md:31 (+ Running Reconnaissance steps) | Read-only mandate is a single parenthetical in a routing bullet. ecs-operation-review's "Access model — READ-ONLY" CAN/CANNOT block + "never mutate; if a check would require a write, mark UNKNOWN" invariant is the family pattern — mirror it where executing agents actually operate. | cf. skills/ecs-operation-review/SKILL.md:59-65 |
| 20 | SKILL.md:167, :222 | `.ecs-recon-report.yaml` lands in cwd with no sensitivity guidance, embedding account ID, subnet/SG IDs, IAM role ARNs, secret ARNs — a dotfile `git add -A` sweeps up. Add: treat as sensitive, don't commit, recommend gitignore; constrain the write location to the workspace (sibling precedent). | — |
| 21 | SKILL.md:123; observability.md:139-143 | Two credential-exposure holes in otherwise disciplined scoping: the quick-reference `describe-task-definition` has no `--query` (raw response includes `containerDefinitions[].environment` — plaintext creds in real estates), and observability pulls the full `logConfiguration` (FireLens options carry API keys). The report schemas exclude these; agent context doesn't. Scope the quick-ref like task-definitions.md:90-107; add "never reproduce `environment` values or non-awslogs `logConfiguration.options`." | — |
| 22 | compute.md:23-27 | IAM prerequisites list only `ecs:DescribeClusters`/`ecs:DescribeServices` but the module calls `describe-capacity-providers` — a policy provisioned from the prereqs gets AccessDenied mid-module. Add `ecs:DescribeCapacityProviders` (all other modules' IAM lists verify clean). | Service Authorization Reference (ecs) |
| 23 | SKILL.md:3 (description) | Three description-level routing issues: (a) "review a service" collides verbatim with ecs-operation-review's "review my ECS services" trigger — the one leaky word in an otherwise crisp score-vs-discover boundary; use "describe/inspect"; (b) "Skip for deployment and CI/CD (ecs-devops)" contradicts the same description's "Detects... deployment configuration... IaC/CI-CD tooling" — qualify as "deployment strategy design and CI/CD engineering"; (c) nonexistent siblings `ecs-cost-intelligence`/`ecs-modernize` unhedged in the description (body line 26 too) — add "(once available)" per family convention. | — |
| 24 | iac.md:44, :256-257, :299, :365; cicd.md:272-287, :334-341 | Detection-confidence framing: "IaC tools consistently tag the resources they manage" is false for Terraform (no default tags — a fully-Terraformed estate lands as `undetermined`, glossed as "no tool detected"); generic stack-name patterns (`<app>-<env>-<svc>`) yield HIGH confidence; cicd's "known tag patterns" are team conventions presented as tool-emitted, and cicd's schema has no confidence field at all (iac's does). State that Terraform is expected to be invisible to tag-based detection; cap name-pattern confidence at medium; add a confidence field to cicd mirroring iac. | — |
| 25 | SKILL.md:222, :230-235, :266-381 | Report/summary framing: single-region, services-only output titled "ECS Reconnaissance Report" with no coverage/limitations metadata, and Step 6 sets no guardrail on summary claims — invites "environment is well-configured" verdicts from a skill that must not score. Also compute.md:230's "running == desired and pending == 0 → service is healthy and at target" is a health verdict (crash-looping services match); reframe as "task counts at target." Add a `coverage:` metadata block + "summarize facts, not verdicts." | — |
| 26 | autoscaling.md:380-392 + schema | The file names the scheduled-scaling-only state but never calls `describe-scheduled-actions`, so such services report `configured: true, scaling_policies: []`, hiding active time-based scaling; suspended targets likewise report identically to active ones (no `suspended_state` field despite the edge case at :394-404). One extra command + two schema fields. | [API_DescribeScheduledActions](https://docs.aws.amazon.com/autoscaling/application/APIReference/API_DescribeScheduledActions.html) |

### NUANCED (clarify)

| # | File | Issue | Both sides |
|---|------|-------|-----------|
| 27 | networking.md:98 | "networkMode defaults to bridge on EC2" — the API's own first sentence says this unqualified, but Windows EC2 tasks default to `<default>` (NAT) and bridge isn't valid there. Qualify Linux-only. | [API_TaskDefinition networkMode](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskDefinition.html) |
| 28 | security.md:326-328 | ECS Exec prerequisites: on EC2, "if no task role is configured, the instance role... is used instead" — the flat "task role must have SSM permissions" would misreport Exec as non-functional for such tasks. | [ecs-exec.html](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html) |
| 29 | networking.md:41-56, :313-361 | App Mesh detected with no lifecycle note: end of support 2026-09-30 (~11 weeks out; new-customer onboarding closed since 2024-09-24; existing customers functional today). Detection mechanics stay valid — but a recon report flagging `app_mesh: true` without the EOS date omits the most consequential fact about that finding. | [what-is-app-mesh](https://docs.aws.amazon.com/app-mesh/latest/userguide/what-is-app-mesh.html) |
| 30 | SKILL.md:125 | "Fargate profiles — N/A" row: Fargate profiles are an EKS-only concept; the row is template residue (the only EKS leak found — everything else greps clean). Delete it. | — |
| 31 | overview.md:323 vs :327; compute.md example :165 | `launch_type: null` means both "capacity-provider strategy in use" (:323) and "skipped for performance" (:327) — undecidable in the report; compute.md uses `not_applicable` for the same fact. Align on `not_applicable`, reserve null for not-collected. (Minor: compute.md:165 example shows `capacityProviderStrategy: []` where the API returns `null`.) | — |
| 32 | security.md:236 | `ecs_exec_enabled: bool` asserts a capability the field doesn't prove (three prerequisites live only in prose). Rename `enable_execute_command_flag` or add a per-field caveat. | — |

### Advisory (post-merge backlog — none blocking)

- Deployment history APIs: `ListServiceDeployments`/`DescribeServiceDeployments`/`DescribeServiceRevisions` (Nov 2024) are the only surface where native blue/green rollback states (`ROLLBACK_*`) are visible — natural follow-up to must-fix #2.
- `schedulingStrategy` (REPLICA|DAEMON) never collected — DAEMON task counts mislead without it.
- `ListAccountSettings` unused (containerInsights account default — a cluster inheriting the default is reported correctly only by luck; Fargate FIPS; dual-stack).
- No storage dimension (task-def volumes EFS/FSx, service `volumeConfigurations` EBS, ephemeral sizing).
- No ECR cross-check (scan config, tag immutability) for enumerated images — if that's ecs-security's lane, say so in Scope Boundaries.
- No subagent orchestration mode (eks-recon makes it a MUST with 9 agents/*.md; 10 modules × drill-down has worse context bloat than EKS) — at minimum state the inline-only decision.
- Multi-region: report is silently single-region; a coverage note + "other regions?" prompt would prevent partial pictures (pairs with #25's metadata block).
- cicd.md:192-196 CodeDeploy detection is account-wide with no cluster-scoping instruction (the CodePipeline step has one) — cross-cluster misattribution risk.
- networking.md:441-446 `load_balancers: []` framed as "no external routing" — misses external target-group registration, VPC Lattice, public task IPs; soften to "no ECS-managed LB association."
- SKILL.md duplication: ~140 lines (Quick Detection Reference CLI table, region resolution, 115-line example report) restate reference content — the quick-ref invites skipping references, which SKILL.md:156 itself forbids. Trim.
- Boilerplate drift: overview.md lacks the TOC the other 9 have; throttle/no-retry policy stated only in overview.md; observability.md:263's "Docker default json-file" is EC2-only (Fargate: logs unretrievable).
- eks-recon's description doesn't exclude ECS reciprocally, and no merged ecs-* sibling routes discovery out to ecs-recon — same fan-out pattern as #133; can fold into that issue.
- Description lacks the body's strongest passive trigger ("run before any ECS change needing current-state context" — cf. eks-recon's "when in doubt, run recon first").

### Strengths

- Read-only contract: 19/19 distinct AWS calls are List/Describe/Get; secrets by ARN/name only; boundary discipline vs ecs-operation-review/ecs-architect genuinely good ("report factually without judgment" appears in the modules).
- Structural/CI: zero findings — already rebased onto main, passes the new strict frontmatter validator ("28 skills validated, 0 errors"), docs-regen parity green, manifest/tables consistent, mirror parity clean, no internal content.
- Frontmatter: YAML valid, description 882 chars — both family classics avoided first-time.
- Verified-correct core (sample): role-split explanation, valueFrom SM-vs-SSM classification, readonlyRootFilesystem-breaks-Exec, Container Insights enum + absent→account-default, predefined scaling metric names, deployment status enum, ECS vs CodeDeployToECS provider mapping, circuit-breaker-is-ECS-rolling-only, `aws deploy` CLI naming, elbv2 casing, launchType/CP-strategy mutual exclusion.
- Pagination `failures[]`-array handling, throttle-no-retry policy, and the autoscaling module's "AccessDenied is not `configured: false`" are exactly the report-shape rigor this family needs.
- JP round-1 fixes: all four verified correctly landed (Service Connect PRIMARY-deployment read at all sites; 4-value launchType enum; routing; ECS Exec requirements).

utkarpun (approved):
## Fix verification — round 2 findings: all addressed ✅

Verified all 32 round-2 findings against the 3 fix commits (0323ffd, a2deaef, 072c528), per-finding with quoted evidence; new factual claims introduced by the fixes were re-verified against live AWS docs (linearConfiguration/canaryConfiguration, CapacityProvider type enum, managedInstancesProvider, PredictiveScaling, batching caps, IMPORT_* statuses, AdvancedConfiguration fields — all correct). Frontmatter parses, description 943 ≤ 1024, strict validator passes, docs-sync green.

The sourcing work deserves a callout: 0 → 65 docs links, `## Sources` in all 11 files, 15 dated freshness gates. Thank you for the thorough turnaround.

### Post-merge follow-ups (non-blocking, from the new fix content)

1. **iac.md — `aws:cdk:path` is not a resource tag.** CDK path metadata lands in the CloudFormation template `Metadata` attribute, not as resource tags, and the `aws:` tag prefix is reserved (iac.md:167 quotes this rule — the tag-table row at :173 contradicts it). The `aws-cdk:*` rows (:174) and stack-association signals (:261) are correct; drop the `aws:cdk:path` tag rows/examples (:46, :98, :173, :345, :403, :526, :535-539).
2. **networking.md:507 — NLB blue/green.** Per [API_AdvancedConfiguration](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AdvancedConfiguration.html), for an NLB `productionListenerRule` holds a *listener* ARN (token `listener`, 4 path segments), so the listener-rule derivation only matches ALB and NLB services degrade to `type: "unknown"` unnecessarily. Add the listener-ARN variant branch.
3. **networking.md — new blue/green content unsourced.** :154/:496-510 lacks a Sources entry (AdvancedConfiguration) and a dated gate, unlike the rest of this commit.

Approving; merging per team plan with the above tracked as follow-ups.

PR #142, opened 2026-07-17
fix(ecs-recon): follow-ups from #137 review

utkarpun:
Non-blocking follow-ups flagged in the APPROVE review on #137, plus fixes from an adversarial review of this PR's own first commit (second commit).

## Fixes

1. **CDK tag detection — remove `aws:cdk:path` as a tag signal** (`references/iac.md`). CDK's construct path is recorded in the CloudFormation template `Metadata` attribute, not as a resource tag, and the `aws:` tag prefix is reserved for AWS use so no user or third-party tool can apply it as a tag (the file's own verified note at the tag table already states this rule). CDK detection on ECS resources now leads with `aws:cloudformation:*` tags + stack association (CDK-pattern logical IDs, `AWS::CDK::Metadata` resource); `aws-cdk:*` construct tags (e.g. `aws-cdk:auto-delete-objects`) remain a valid signal when present but are emitted on few resource types (e.g. S3 buckets) and rarely appear on ECS resources. Refs: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-resource-tags.html, https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd.html

2. **NLB blue/green load balancer derivation** (`references/networking.md`). Per the ECS API reference, `advancedConfiguration.productionListenerRule` holds a listener-rule ARN for an ALB but a plain listener ARN for an NLB. The derivation previously only handled the `listener-rule` ARN shape, so NLB blue/green services degraded to `type: "unknown"` unnecessarily. Added the `listener` token branch (replace `listener` with `loadbalancer`, keep `net/<lb-name>/<lb-id>`); stale-LB fallback semantics unchanged. Ref: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AdvancedConfiguration.html

3. **Sourcing for blue/green content** (`references/networking.md`). The `advancedConfiguration` content had no source citation or dated freshness gate, unlike the rest of the file. Added API_AdvancedConfiguration.html to the Sources section and "Facts verified 2026-07-17" gates adjacent to the blue/green content, matching the file's existing gate style.

## Second commit (review follow-ups on this PR)

An adversarial review of the first commit surfaced four refinements, applied in the second commit:

- Softened "CloudFormation **always** applies `aws:cloudformation:*` tags" — stack-level tag propagation varies by resource type per the CloudFormation docs; ECS clusters, services, and task definitions do receive them.
- Scoped "nothing can apply it as a tag" to "no user or third-party tool" (AWS services themselves do apply `aws:`-prefixed tags).
- Reworked the three ECS-resource examples that depicted `aws-cdk:auto-delete-objects` on an ECS service (that tag is emitted on S3 buckets) to use the realistic CDK-on-ECS evidence: `aws:cloudformation:*` tags + stack association.
- Split the `aws:cdk:path`-in-Metadata claim into its own dated gate citing the CDK CLI reference (the previous gate's CloudFormation URL didn't cover it), and fixed an ALB-only wording in networking.md's target-group attachment description ("listener rules (ALB) or listeners (NLB)").

## Also included (minor)

- `SKILL.md` module table: qualified "account-wide inventory of clusters, services, tasks" to match the module's `scope_note` (standalone `run-task` / scheduled tasks are not discovered).
- `references/overview.md`: added a dated freshness gate for the DescribeClusters 100-name cap (verified 2026-07-17 against https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeClusters.html).

Docusaurus mirrors regenerated via `misc/update-all-references.sh` + `misc/update-pages.sh`; `validate-frontmatter.py` passes (28 skills, 0 errors).

### DIFF
--- a/misc/website/docs/skills/ecs/ecs-recon/index.md
+++ b/misc/website/docs/skills/ecs/ecs-recon/index.md
@@ -97,7 +97,7 @@ Load only the references needed for the user's request — this keeps context fo
 
 | Module | Intent / When to Use | Reference File |
 |--------|---------------------|----------------|
-| Overview | Always loaded first — account-wide inventory of clusters, services, tasks | [overview.md](references/overview) |
+| Overview | Always loaded first — account-wide inventory of clusters, services, and their tasks (standalone/scheduled tasks not discovered) | [overview.md](references/overview) |
 | Compute | Capacity providers, launch types, task counts, Fargate vs EC2 | [compute.md](references/compute) |
 | Task Definitions | Container images, CPU/memory, family and revision | [task-definitions.md](references/task-definitions) |
 | Deployment | Rolling update, CodeDeploy, circuit breaker, min/max percent | [deployment.md](references/deployment) |
--- a/misc/website/docs/skills/ecs/ecs-recon/references/iac.md
+++ b/misc/website/docs/skills/ecs/ecs-recon/references/iac.md
@@ -54,7 +54,7 @@ Run detections in this order — each step adds evidence and confidence:
 ```
 
 **Why this order matters:**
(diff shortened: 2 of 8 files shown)


literalice (approved):
Reviewed as the original ecs-recon author — all three fixes are correct (verified `aws:cdk:path` against the CDK CLI docs and the ALB/NLB `productionListenerRule` split against API_AdvancedConfiguration). Nice call using `aws:cloudformation:*` + stack-association for CDK-on-ECS. LGTM 👍

PR #152, opened 2026-07-21
fix(ecs-recon): correct aws:cdk:path claim + NLB blue/green LB derivation

literalice:
## Summary

Follow-up to #137. That PR merged with three small residual items flagged in the final review round (all in already-merged content, **none are regressions**). This PR addresses them.

### `aws:cdk:path` is not a resource tag (`iac.md`)
`aws:cdk:path` is written into the **CloudFormation template's per-resource `Metadata` section**, not applied as a resource tag — so it never appears in `ecs:ListTagsForResource` output. Classifying it as a tag was both incorrect and self-contradictory with the file's own "`aws:` prefix is reserved for AWS use" note.

- Removed `aws:cdk:path` from the tag-pattern table, tool-mapping table, the CDK tag example, and the overlapping-evidence edge case.
- Kept the real `aws-cdk:*` construct tags (e.g. `aws-cdk:auto-delete-objects`).
- CDK detection now routes through **stack association** (CDK-pattern logical IDs / `CDKMetadata` resource), which the module already documents.

### NLB blue/green load balancer derivation (`networking.md`)
`advancedConfiguration.productionListenerRule` holds a **listener-rule ARN for an ALB** but a **listener ARN for an NLB** (NLBs have no listener rules). The previous ALB-only derivation silently degraded NLB blue/green services to `type: "unknown"`.

- Documented both ARN shapes and a single derivation (strip to the `loadbalancer/<type>/<name>/<id>` triplet) that covers `app` and `net`.

### Sourcing for the blue/green content (`networking.md`)
Added `## Sources` entries (`API_AdvancedConfiguration`, NLB blue/green resources) and a dated freshness gate for the new blue/green edge case, matching the family convention applied throughout the rest of the skill.

## Verification

- All claims verified against AWS docs (2026-07-14): [API_AdvancedConfiguration](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AdvancedConfiguration.html) (ALB→listener-rule / NLB→listener), [NLB blue/green resources](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/nlb-resources-for-blue-green.html), and CDK metadata semantics.
- Docs regenerated via `update-all-references.sh` + `update-pages.sh`; frontmatter validates (29 skills, 0 errors); mirror parity clean.

- [x] `skills/ecs-recon/SKILL.md` present with valid frontmatter (unchanged this PR)
- [x] Ran the docs regeneration scripts and committed the resulting mirror changes
- [ ] `misc/evals/ecs-recon/triggering.json` — not included (consistent with #137 and the current ECS skill precedent)

### DIFF
--- a/misc/website/docs/skills/ecs/ecs-recon/references/iac.md
+++ b/misc/website/docs/skills/ecs/ecs-recon/references/iac.md
@@ -54,7 +54,7 @@ Run detections in this order — each step adds evidence and confidence:
 ```
 
 **Why this order matters:**
-- Tags are the strongest single-call signal for the tools that DO tag: CloudFormation always applies `aws:cloudformation:*` tags, CDK adds `aws:cdk:path` (when metadata is enabled), and Copilot adds `copilot-*` tags
+- Tags are the strongest single-call signal for the tools that DO tag: CloudFormation always applies `aws:cloudformation:*` tags, CDK may add `aws-cdk:*` tags on some constructs, and Copilot adds `copilot-*` tags. (Note: `aws:cdk:path` is **not** a resource tag — CDK writes it into the CloudFormation template's per-resource `Metadata` section, so it never appears in `ListTagsForResource` output. Detect CDK via stack association instead — see step 2.)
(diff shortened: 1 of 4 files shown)


literalice:
Closing in favor of #142, which was opened earlier by @utkarpun and covers the same three items (aws:cdk:path tag correction, NLB blue/green LB derivation, blue/green sourcing) — plus additional refinements. I didn't spot #142 before opening this. Reviewing #142 instead.

PR #166, opened 2026-08-05
feat(graviton-migration): add first-party Graviton/arm64 migration skill

utkarpun:
## What

Adds a new first-party **`graviton-migration`** skill (general group) that owns the *execution* of an x86→AWS Graviton (arm64) workload migration. The existing catalog treats Graviton as a cost lever (eks-cost-intelligence, eks-best-practices); this skill does the actual move.

## Contents

- **`SKILL.md`** — sets up the third-party Arm migration MCP server (`armlimited/arm-mcp`) for arm64-readiness scanning, then orchestrates the migration workflow. The MCP setup is **harness-aware**: verified config for Claude Code (`${CLAUDE_PROJECT_DIR:-.}`) and Kiro (`${workspaceFolder}`), plus a self-correcting guard that tells the agent to web-search (not recite from memory) the MCP-config format for any other harness — the stable invariant is the `docker run … armlimited/arm-mcp` server entry; the config wrapper varies. Source mounted read-only.
- **`references/scanner-workflow.md`** — pre-migration arm64-readiness scan (native deps, arch-specific binaries, base-image arch variants, per-language runtime rules).
- **`references/karpenter-migration.md`** — taint-first Karpenter **v1** arm64 NodePool cutover runbook (canary toleration+nodeSelector, DaemonSet arch caveat, progressive shift, arch spread).
- **`references/multi-arch-pipelines.md`** — per-CI multi-arch (amd64+arm64) build matrices (GitHub Actions, CodeBuild, GitLab, Jenkins) + manifest-list explanation.
- Eval scaffold (`triggering.json`, `evals.json`, `.skilleval.yaml`, README SIBLING_MAP) with **bidirectional** sibling exclusions vs eks-cost-intelligence, eks-best-practices, eks-mcp-server, plus attributions for eks-ingress-migration / eks-upgrade-check.
- Regenerated README/skills.json/docusaurus mirrors.

## Boundary

`graviton-migration` = migration **execution** (readiness scan → NodePool cutover → multi-arch CI). Cost-savings scoring / adoption % → `eks-cost-intelligence`; "should we adopt Graviton?" advisory → `eks-best-practices`; the *EKS* MCP server → `eks-mcp-server`.

## Validation

- `validate-frontmatter.py`, `update-all-references.sh --check`, `update-pages.sh --check`, and `check_hygiene.py --skill graviton-migration` all pass.
- Reviewed adversarially (multi-pass, verification-gated) to a clean final pass. Facts spot-checked live: `armlimited/arm-mcp` image exists on Docker Hub (multi-arch); `${CLAUDE_PROJECT_DIR:-.}` default requirement confirmed against Claude Code MCP docs; Karpenter v1 API and CI action versions verified.
- Dry-run exercised against a synthetic Kong-on-EKS + Netty-native-classifier + Karpenter-v1 + DaemonSet shape: caught the x86-native-classifier blocker, the DaemonSet taint-bypass, produced the taint-first canary, and surfaced its own out-of-scope gaps (JDK-version upgrade, node-AMI, IaC drift) rather than skipping them.

## Notes

- First-party authored (no upstream vendor sync): no `.vendor-sha`, no `sync-*.sh`, no `vendor-update.yml` entry.
- Lives in `skills/` only — not ported to `devops-agent/` (that runtime rejects MCP refs).


utkarpun:
Updated: rebased onto current `main` and pushed a single clean commit. The push is based on the latest `main`, so the diff is graviton-migration additions plus three reciprocal routing exclusions only — no changes to other skills' content.

**Cutover runbook reworked.** The Karpenter cutover now leads with a **single multi-arch Deployment + rolling update** as the default, with `kubectl rollout undo` as the abort (it reverts the pod template but not the replica count, so an HPA can't fight the rollback). The **two-per-arch-Deployment** pattern is kept as an explicitly-flagged advanced option for a tunable percentage soak, with its abort corrected to pause/delete the per-arch HPA before shifting replicas. Added an EC2NodeClass arm64-AMI verification step (arch-aware `alias` resolves arm64 automatically; an x86-pinned `amiSelectorTerms` leaves the NodeClaim unfulfilled), and made the arm64 NodePool on-demand-only during the canary/soak window (Spot reclaims otherwise corrupt the validation signal).

**Validated end to end:**
- Scan path: drove the Arm migration MCP server for real — tool set, `migrate_ease_scan` usage, scanner/arch enums, the layer-3 ELF check, and the gate logic all behave as documented. The knowledge-base search is a local bundled index (confirmed offline).
- Cutover path: exercised the default rolling cutover and the `rollout undo` abort against a live EKS + Karpenter cluster on a throwaway workload — an arm64 node provisioned from an arch-aware AMI alias, the rolling cutover kept the Service serving, and the abort reverted cleanly. All test resources were torn down.

Fact corrections in this revision: neoverse-512tvb GCC floor (11), .NET Core arm64 Linux start (3.0), zlib-ng arm64 support (present since 1.x), and the knowledge-base tool described as local rather than networked.

PR #177, opened 2026-08-13
Add the ecs-modernize skill: assess an existing app, then execute a gated migration onto Amazon ECS

literalice:
## Summary

Adds **`ecs-modernize`**: assess an existing application (VMware / EC2) by source-code analysis, decide between Replatform and Rearchitect, and execute the approved migration onto Amazon ECS.

The skill is a two-phase pipeline with an explicit control point between the halves. The **Assessment_Phase** is strictly read-only — static source analysis, a weighted 0–100 Fit_Score over six dimensions, and a strategy recommendation — and its only side effect is one Markdown report. **Migration_Execution** never starts until the user explicitly approves a strategy and target path at the Execution_Gate, and each action class (image push, file generation, `terraform apply`) then asks for its own separate confirmation.

Also included: a hands-on exercise, evals, and the `dotnet-aws-ecs` vendoring the shared generated surfaces depend on.

## What the skill contains

12 reference modules routed by user intent, so a single-module request loads only its transitive prerequisites rather than the whole corpus:

- **Assessment** — tech-stack detection (including the Liberty-vs-Tomcat `server.xml` collision resolved by content, not filename), blocker detection across six categories with a two-class remediation model, Fit_Score scoring, both path modules, and report generation.
- **Execution** — code transformation (AWS Transform orchestration plus agent-executed items), containerization, environment build, and deploy with steady-state verification.
- **WebSphere support** — traditional WAS containerized as-is on IBM's ILAN-licensed image, or migrated to Liberty as a Rearchitect item, with IBM's assessment tooling surfaced as accelerators.

## The delegation boundary, and why it is not uniform

Environment IaC generation splits by target shape, which is worth flagging because the asymmetry is deliberate:

- **Rearchitect compute models** (Express Mode / Fargate / Managed Instances) are delegated to `ecs-build`.
- **Replatform** (ECS on EC2, containerize-as-is) and the **Windows container path** are generated in-skill.

`ecs-build` generates `network_mode = "awsvpc"` task definitions exclusively, and services that always carry a `capacity_provider_strategy`. It has no notion of ALB target-group stickiness, `bridge` networking, or dynamic host port mapping. Those four elements *are* the Replatform shape, so delegating that path would produce an environment contradicting the approved strategy. `SKILL.md` carries a coverage table stating what would have to change in `ecs-build` for each carve-out to become a delegation again.

The delegated half was verified by running `ecs-build` against this fixture's handoff inputs: 28 validator checks passed, and it correctly omitted the stickiness block because the handoff said no session affinity was needed.

**One change to `ecs-build` in this PR:** a routing-description carve-out so it excludes the `ecs-modernize` cases (matching the `ecs-architect` and `ecs-recon` convention), with the skills manifest, README table, and website wrapper regenerated. Two further improvements surfaced during that verification and are deliberately left for a separate PR.

To keep the generated Replatform environment predictable, the skill ships a verified skeleton at `assets/replatform-terraform/` and copies it rather than improvising: two runs of the same assessment differ only in `terraform.tfvars`, so a reviewer can diff them. Both network modes are supported — CREATE builds a VPC (the normal case for an on-premises or VMware migration, which has no target VPC), REUSE takes an existing one. Requiring an existing VPC would have blocked the most common lift-and-shift starting point.

## The exercise

`examples/ecs-modernize/liberty-to-ecs/` — a Java 8 / Java EE 7 Liberty application with 11 planted blockers across all six scoring dimensions, plus a remediated reference implementation. The loop is assess → decide → execute → re-assess.

Two deliberate choices:

- **No wrapper scripts.** Every Terraform, Docker, Maven and AWS CLI step is spelled out, because wrapping them hides exactly what a hands-on exists to show. One script remains, generating the `system`-scoped JAR so the repo stays source-only.
- **The agent does the work.** Part 2 has it generate the Replatform environment; Part 3 has it transform the code (measured: 5 work items, ~15 min, including one expected build failure and its in-branch fix) rather than deploying a pre-made result.

Scores in the README are **measured, not calculated**: 23 before, 85 after. An earlier hand-calculated table was wrong on four of six dimensions — most notably `framework_modernity`, estimated at 70 when Jakarta EE sits on tier 3 of the modernity ladder and cannot exceed 59.

## Verification

The skeleton was deployed against a live account in an isolated VPC, then destroyed.

| Check | Result |
|---|---|
| `terraform apply` | 26 added, 0 changed, 0 destroyed |
| Container instances registered | 2, within ~10s |
| Service steady state | desired 2 / running 2 |
| Target registration | ephemeral ports (32768), not 9080 — dynamic mapping visible externally |
| Application | `/orders/health` → `{"status":"UP"}` |
| Stickiness ON (cookie kept) | 10/10 requests same session |
| Stickiness bypassed (AWSALB dropped) | 5/5 split — session invisible to the other task |
| State loss after task replacement | archive 1 file → 0 on both tasks |
| ECS Exec | works; archive inspected directly |
| `terraform destroy` | 26 destroyed; ALB / EC2 / VPC / ECS / ECR / log group all confirmed gone |

Skeleton `terraform validate` passes in all four shapes: CREATE, REUSE, stickiness disabled, and skeleton-plus-EFS-fragment. Both applications build and run; the assessment and the code transformation were each executed end to end.

### One real defect found by deploying

CREATE mode built public subnets with no NAT gateway while `assign_public_ip` defaulted to `false`, leaving instances with no internet path — the ECS agent could not reach the control plane or pull from ECR. **`terraform apply` succeeded; only steady state failed**, which is the failure mode static validation cannot catch. Public IPs are now implied by CREATE mode, and the REUSE-mode variable documents that a wrong value fails silently the same way.

Two README errors were also corrected: the ECS Exec container name, and the session-loss demonstration silently appearing to *pass* when only one target is healthy — an ALB routes only to healthy targets, so the loop cannot reach the other task. Measured both ways and documented the check.

## Notes for reviewers

- Every external citation was resolved against live docs. Three IBM URLs in an earlier draft were fabricated slugs that returned HTTP 200 — IBM Docs serves a generic page for any invalid `?topic=`, so a status check alone does not catch them. They were replaced with topics verified by title.
- `dotnet-aws-ecs` vendoring follows the existing convention (`LICENSE` + `UPSTREAM.md` + a `misc/sync-*.sh` script + a `THIRD_PARTY_NOTICES.md` entry) and is included because the shared generated surfaces carry its entries.
- An unrelated `ecs-recon` fix that had been sitting on the same branch was rebased out and will go up separately.
- `validate-frontmatter.py`: 32 skills, 0 errors. `update-all-references.sh --check`: clean under `C.UTF-8`, the runner locale.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)


utkarpun (dismissed):
Thanks for the continued work — re-reviewed against the current head (`19d41c3`). Before the findings, two **corrections to my own prior review**, in fairness: (1) I earlier said the WebSphere 9.0 session-replication claim "could not be faulted" — that was wrong, it's a must-fix below; (2) I flagged the IBM `?topic=` URLs as unverifiable — they're actually real (the 403s were bot-blocking), so disregard that. I ran the full factual/security/determinism pass over the whole branch again; the two-phase Assessment→Execution_Gate design, the by-design fixture credentials, the weight/band partition, and the worked Fit_Score example all still hold up.

Note on `19d41c3`: its text edits (pointing the Liberty exercise at the scripts it has) are correct and well-targeted — but see the hygiene item under should-fix; the "comments only" message isn't accurate.

## Must-fix

**1. `iam-policy.json` `ExecutionTerraformIAM` is privilege-escalation, not "least-privilege."**
`references/iam-policy.json` (`ExecutionTerraformIAM`) grants `iam:CreateRole` + `iam:PutRolePolicy` + `iam:AttachRolePolicy` + `iam:PassRole` (+ `CreateInstanceProfile`/`AddRoleToInstanceProfile`) on `Resource: "*"` with no `Condition`. That combination is admin-equivalent (create a role → attach `AdministratorAccess` → pass it). `SKILL.md:233` calls the document "least-privilege," which is misleading. Scope `PassRole` with `iam:PassedToService` (`ecs-tasks.amazonaws.com`, `ec2.amazonaws.com`), constrain `AttachRolePolicy` to a role path / allow-list (or a `PermissionsBoundary`), or drop the "least-privilege" wording. ([PassRole scoping guidance](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create-considerations.html))

**2. Generated `awslogs` config omits `mode` → silent log loss during the verify window.**
`assets/replatform-terraform/main.tf` (logConfiguration.options, ~L339-344) sets group/region/stream-prefix but no `mode`. Since [ECS changed the awslogs default from `blocking` to `non-blocking` on June 25, 2025](https://repost.aws/knowledge-center/ecs-eks-troubleshoot-container-logs), an unset `mode` now defaults to non-blocking and drops logs under buffer pressure — undercutting the skill's own steady-state failure diagnosis in `deploy-verify-handoff.md`. Set `mode` (and `max-buffer-size`) explicitly. ([Preventing log loss with non-blocking mode](https://aws.amazon.com/blogs/containers/preventing-log-loss-with-non-blocking-mode-in-the-awslogs-container-log-driver/))

**3. Eval id 2 (`struts-orders`) asserts a recommendation the rubric doesn't produce.**
`misc/evals/ecs-modernize/evals.json` id 2 expects Fit_Score < 40 → Replatform. Re-deriving by hand against `scoring-and-recommendation.md`: a clean Tomcat servlet app floors `os_host_dependency` (L138) and `dependency_coupling` (L215-216) in the 60-89 band, and all three blockers are `replatform_ok` (zero `must_fix`). That lands **~44 → grey zone (40-69, "present both strategies")**, not sub-40. Sub-40 is only reachable by pinning both clean dimensions to their band floor, which the "position by scope and severity" mandate (L100) doesn't justify. Either adjust the fixture to something that genuinely scores <40 (e.g. add a `must_fix` blocker or real vendor/proprietary coupling), or change the expected outcome to grey-zone/parallel.

**4. WebSphere 9.0 "removed memory-to-memory replication" is incorrect.**
`replatform-path.md:194` states tWAS 9.0 removed memory-to-memory session replication "in favor of eXtreme Scale." IBM's WAS-ND 9.0.5 docs still document it: *"Select **Configure HTTP session memory-to-memory replication**… the web container for each cluster member is configured for memory-to-memory replication."* ([Creating clusters, WAS-ND 9.0.5](https://www.ibm.com/docs/en/was-nd/9.0.5?topic=workloads-creating-clusters)) DRS memory-to-memory replication is retained in 9.0 — please correct the claim.

**5. `for_each` over unknown subnet IDs breaks EFS+CREATE at `terraform plan`; "validates in all four shapes" overstates coverage.**
`assets/replatform-terraform/OPTIONAL-efs.tf.fragment` (~L56) keys `aws_efs_mount_target.archive` on `for_each = toset(local.subnet_ids)`. In CREATE mode `local.subnet_ids = aws_subnet.public[*].id` (`main.tf:51`) — unknown until apply — so `terraform plan` hard-errors `Invalid for_each argument … cannot be determined until apply` (reproduced). It works only in REUSE mode. Because `terraform validate` is value-independent, "validates in all four shapes" passes while EFS-on-a-fresh-VPC — a plausible VMware-migration default — can't even plan. Use `count`-based mount targets, or document the fragment as REUSE-mode-only, and reword the coverage claim.

**6. Japanese draft markers left in the shipped skill, including emit-visible positions.**
Non-English tokens remain in `deploy-verify-handoff.md` (L62, 166, 213, 214, 250, 287), `containerization-execution.md:352`, and `code-transformation.md:238` (also propagated to the website mirror). Two are model-visible, not inert: `deploy-verify-handoff.md:213-214` places `未確定` inside the YAML handoff **schema example** (`value: string | 未確定`), and `code-transformation.md:238` instructs *saying* `能力差の証拠なし`. A model following those will emit the marker into real output. Please replace with English equivalents.

## Should-fix

- **Eval id 8 (`binaries-only`) `fit_score: null` collides with the 0-29 "No build definition" band.** The rubric's `build_reproducibility` 0-29 band reads "No build definition" (L237), which invites a determinable ~10 score, while the edge-case note (L329-330) distinguishes "checked, found none" (scored) from "could not check" (undetermined). Binaries-with-no-build satisfies both readings, so a faithful run may produce a numeric score rather than `null`. Add an explicit rule for compiled-artifact-only input.
- **`dotnet-aws-ecs` and `ecs-modernize` have no bidirectional routing exclusion.** `dotnet-aws-ecs/SKILL.md:3` targets ".NET → ECS Fargate on Linux + ALB," which overlaps `ecs-modernize`'s Rearchitect .NET target; neither skill's Skip / "Don't use" list names the other, so a ".NET to ECS Fargate" query is genuinely ambiguous. Add mutual exclusions. Relatedly, `ecs-build`'s migration trigger has no reciprocal pointer back to `ecs-modernize`, so a "generate Terraform to migrate my app" request can bypass the Assessment/Execution gate.
- **The `dotnet-aws-ecs` "dependency" framing is backwards.** The PR calls it "the vendoring the shared generated surfaces depend on," but nothing under `ecs-modernize/` references it (.NET/Windows are handled in-skill). If it's load-bearing, wire the link; otherwise present it as a standalone skill and justify the ~2,000 vendored lines on their own terms. (Also note `UPSTREAM.md` sources it from a personal repo, `adisimon217/sample-appmod-skills` — worth confirming that's intended for an aws-samples repo.)
- **Shipped skeleton security posture + "verified" framing.** The skeleton is described as a "verified skeleton" (`SKILL.md:224`) but ships HTTP:80-only (no TLS/443, `main.tf:226-235`), `enable_execute_command = true` by default with no `execute_command_configuration` audit/KMS (`variables.tf:200-208`, `main.tf:243-246`), egress `0.0.0.0/0`, and no `health_check_grace_period_seconds` (despite `replatform-path.md` noting legacy apps start slowly — ECS can kill a still-starting task). "Verified" here means `validate` + one deploy, not production-hardened. Add a "before production: TLS, hardening, Container Insights" note (or hand off to `ecs-security`) so the skeleton isn't copied as a hardened baseline, default Exec to `false` or wire logging, and emit a tunable grace period.
- **Modernity ladder lists EOL runtimes/frameworks as "modern."** `scoring-and-recommendation.md:186` puts Spring Boot 2.x and .NET 6 in the 60-79 (≥ modern floor) band, contradicting the EOL→legacy cap at L171. The runtime-constraint rule (L195) rescues .NET 6/7 but not EOL Spring Boot 2.x. Split supported (.NET 8) from EOL.
- **ECS Managed Instances Region exclusion is understated.** `rearchitect-path.md:122` says "Not available in the China Regions"; the [FAQ](https://aws.amazon.com/ecs/managed-instances/faqs/) excludes **both GovCloud (US) and China**. Add GovCloud.
- **Commit `19d41c3` hygiene.** The message says "Comments only — no behaviour change," but its only *added* file is a functional `.terraform.lock.hcl` (pins aws 6.59.0) — and it lands in `examples/ecs-modernize/dotnet-framework-to-ecs/`, an **orphan dir** whose only content is that lockfile (no `.tf`/README, unregistered in `examples.json`). Either remove the stray lockfile or flesh out the example, and keep lockfile hygiene consistent with the Liberty example (which gitignores its lock).

## Advisory (non-blocking)

- **IMDSv2 not enforced** in the skeleton launch template (`http_tokens` defaults to `optional`). The container-credential-theft path is *not* actually open here — default `http_put_response_hop_limit = 1` blocks bridge-networked containers from IMDS — so this is host-level defense-in-depth only; still worth setting `metadata_options { http_tokens = "required" }`.
- **Coverage gaps vs AWS guidance/siblings:** Container Insights (zero mentions), a "should this even move to ECS?" off-ramp (the lowest band still recommends Replatform), and blue/green deployment (only canary via Express Mode).
- **Managed Instances GA dates** ("GA Sept 2025 / all commercial Regions Oct 2025", `rearchitect-path.md:114`) — I couldn't independently confirm these from What's New (search blocked); please attach the What's New posts and phrase "as of <date>".
- **`dotnet-aws-ecs` stale-prone content:** hardcoded pricing tables (`deployment-guide-template.md:216-237`, `cloudformation-template-guide.md:636`) and version matrices would be better as links; acceptable given it's vendored/mirror-bannered.
- **Sync-script robustness:** `misc/sync-dotnet-aws-ecs-skill.sh` runs `cp references/*.md` and the version guard *after* the `rm -rf` (Step 3), so a malformed upstream (SKILL.md present but empty `references/` or missing `version:`) wipes local then aborts; the guard only covers SKILL.md. Minor shellcheck: SC2064 (L67), SC2295 (L212).
- The **"28 validator checks passed"** claim for the delegated `ecs-build` half isn't reproducible from this PR (no output/handoff file shipped) — fine to leave, flagging for awareness.
- `SKILL.md` description is 1009/1024 chars — passes, but thin headroom given the CI hard-fail cap.

## Checked and could not fault

- Read-only Assessment invariant and the Execution_Gate are genuinely enforced in `SKILL.md` (per-action-class confirmations, single-write invariant), not just narrated.
- Fit_Score weights sum to 100, bands partition 0-100 exactly (edges 39/40, 69/70 clean), the 72.35→72 worked example and eval id 1 re-derive correctly, the framework-modernity cap is consistent.
- Fixture credentials are clearly by-design (RFC 2606 domains, labeled PLANTED) — not leaks.
- `terraform validate`/`fmt` pass, the sync script fails non-zero on missing/malformed input (guard fires pre-`rm` for the SKILL.md case), website mirrors are in sync, CI validators are green.
- The IBM `?topic=` documentation URLs resolve to real pages, and the Managed Instances Bottlerocket-only / 14-21-day drain-replace / awsvpc-host claims and the Tomcat/Liberty `server.xml` collision handling are all accurate.

Happy to re-review once the must-fix items are addressed.

utkarpun (commented):
Reviewed the full branch and pushed the fixes directly as append-only commits on top of your work. Nothing was rebased or rewritten; your commit `793816a` is untouched at the base of the stack, with 15 fix commits layered above it.

**Fixes applied**

- Aligned evals 2 and 6 with the scoring rubric, made the eval premises realizable under the runner, and added a `dotnet-aws-ecs` triggering negative.
- Corrected factual and citation points: the `awslogs` default, the WebSphere machine-to-machine citation, the WS2025 base image, and the stale Managed Instances GovCloud exclusion.
- Made the IAM policy actually apply the skeleton and sharpened the privilege-escalation disclosure, so the policy's ability to modify account-wide resources is called out explicitly rather than left implied. The CodeBuild service role is also disclosed.
- Added an `ecs-build` routing carve-out so it hands the `ecs-modernize` cases back, matching the `ecs-architect` and `ecs-recon` convention, with the manifest, README table, and website wrapper regenerated.
- Windows AD/gMSA and FSx disclosures, capacity-provider reversibility, and secrets and decision-table gaps.
- A REUSE-mode ALB precondition for empty subnets, and a website generator fix that was bleeding a section into the next heading.
- Made the ALB health-check matcher configurable (`health_check_matcher`, default `"200"`). It was hardcoded to `200` while `health_check_path` is a variable and the guidance endorses a root-path fallback; a legacy stack answering the root with a redirect or auth challenge (302/401) would otherwise never reach a steady healthy state. The default preserves existing behavior.

**Verified as correct, no change**

- "Spring Boot 2.x left OSS support in November 2023" is accurate. Spring extended OSS support and shipped 2.7.18, the final 2.x OSS release, on 23 November 2023. endoflife.date lists an earlier un-extended baseline; the Spring announcement is authoritative.
- Routing and handoff to `ecs-build`, the vendoring convention for `dotnet-aws-ecs`, and the IAM escalation (disclosed, and intrinsic to the apply path) all hold.

The branch is clean on my side. Two `ecs-build` improvements that surfaced during verification, plus a de-vendoring follow-up for `dotnet-aws-ecs`, are tracked separately and intentionally out of scope here.

PR #92, opened 2026-06-23
feat(skills): add eks-ingress-migration skill

kritktongrakroj:
Assess a live EKS cluster's NGINX/Ingress estate and plan migration to Gateway API, the AWS Load Balancer Controller (ALB), or AWS Transform (ATX). Discovers controllers and routes, computes a Migration Difficulty Score (0-100) with a separate re-architecture gate, and generates per-cluster reports plus ready-to-apply manifests. Read-only.

- skills/eks-ingress-migration/  — SKILL.md (no per-skill license; repo MIT-0 governs), references/ (assessment docs + ATX TD + samples), tools/.
- misc/evals/eks-ingress-migration/ — triggering.json (17), evals.json, README.
- steering/eks.md — one Day-2 hub routing row.
- README.md / skills/README.md / misc/website/** — auto-generated catalog refresh (required by docs-sync CI; additive).
- Sibling disambiguation (required by the new-skill workflow): one ingress-migration near-miss negative + one SIBLING_MAP bullet appended to each overlapping neighbour's eval set — eks-recon, eks-operation-review, eks-upgrade-check, eks-best-practices. Minimal +2/-1 (+1) per file, no reformatting of existing prompts.

## If this PR adds or changes a skill

- [ ] Ran `/apex:new-skill` (or walked the equivalent manual steps in `CONTRIBUTING.md`)
- [ ] `skills/<skill>/SKILL.md` present with valid frontmatter
- [ ] `misc/evals/<skill>/triggering.json` authored (≥16 prompts; balanced positives and near-miss negatives)
- [ ] For each neighbour: `misc/evals/<neighbour>/SIBLING_MAP` gained a bullet and its `triggering.json` gained the matching negatives (via `update_sibling_map.py` or hand-edit)
- [ ] Ran the `update-docs` skill and committed any resulting changes (regenerated wrappers/manifest, marker-block updates, prose edits)

See [`CONTRIBUTING.md`](../CONTRIBUTING.md) for the full new-skill workflow.

PR #204, opened 2026-08-30
feat(eks-ingress-migration): DevOps Agent port

### LINKED ISSUE #199: eks-ingress-migration: DevOps Agent port
## Summary

Create a DevOps Agent port of the `eks-ingress-migration` skill, as part of the DevOps Agent ports effort (#111).

## Context

The `eks-ingress-migration` skill was recently hardened and expanded (#129, #153, #154). The port should be based on the current merged skill on `main`, including the LBC Ingress to Gateway API (lbc-migrate) toolkit path added in #154.

## DevOps Agent runtime constraints

The port must conform to the DevOps Agent runtime:

- No `scripts/` directory. The agent runs fully autonomously, so any logic the skill relied on a script for (notably the HTML report generation) needs to be expressed as agent instructions and templates rather than an executable.
- No `allowed-tools` frontmatter.
- No MCP references.
- Use `references/` rather than `steering/`.
- Size budget: at most 6 MB and 100 files.

## Deliverable

- A DevOps Agent formatted copy of the skill placed under the DevOps Agent tree, carrying the current `main` content (including the lbc-migrate toolkit).
- Sync the #129 learnings (do not rate an empty or absent ingress) into the port.
- Reviewer sign-off before the port is uploaded for testing.

kritktongrakroj:
Closes #199.

DevOps Agent port of `eks-ingress-migration`, based on current merged `main` (`94d4b3f`), including the lbc-migrate toolkit path from #154.

`devops-agent/eks-ingress-migration/` — 33 files, 348 KB (budgets: 100 files / 6 MB). Layout follows the existing ports: `SKILL.md` + 11 flat `references/` + `references/iam-policy.json` + `references/porting-notes.md` + `assets/samples/` (16 YAML) + `assets/atx/` (3 MD). No `UPSTREAM.md` — that only applies to the vendored skill; this one is first-party.

### Alignment with the existing ports

I diffed `skills/eks-upgrade-check` against its port to learn the established transformation, then inventoried all five ports before writing anything. The port is deliberately **minimal-delta**: CLI verbs → API action names, ` ```bash ` → plain fences, file-writing → inline rendering, interactive prompts → hard-stop decision tables, sample-code disclaimer appended to the report template.

One layout correction came out of that inventory: **no existing port nests subdirectories under `references/`**, and data files live in top-level `assets/` — a rule the `eks-upgrade-check` port documents itself ("`data/` and `tools/` → `assets/`"). So upstream's `references/samples/` and `references/atx/` became `assets/samples/` and `assets/atx/`.

Hard-stop-on-ambiguity follows `eks-upgrade-check`; `eks-recon` instead assesses all discovered clusters. Both are in-repo precedents — I took the hard stop because this skill emits a large scored report per cluster and renders it inline.

### The HTML report (the one real design question)

`tools/report_to_html.py` cannot run here, so report generation is re-expressed as instructions + the markdown template, and the tool's *deterministic* renders became literal text:

| Upstream token | Port |
|---|---|
| `[[SCORE:nn:LABEL]]` | `**Migration Difficulty Score: <nn> / 100 — <LABEL>**` |
| `[[GATE:n]]` | the tool's own strings, verbatim: `✓ No re-architecture blockers` / `⛔ N blocker(s) need(s) redesign / approval` |
| `[[DL:*]]` download buttons | cross-references to the Export Materials section, where the YAML is rendered inline |
| `!!hot!!` | `**bold**` |

This is not cosmetic. `report_to_html.py`'s `inline()` implements only `!!…!!`, `**…**`, backticks and links — it has no markdown fallback — so every one of those tokens would otherwise reach a markdown-only reader **literally**.

I deliberately did **not** hand-template an HTML/JS bundle. It could not be fixture-tested, and the port states plainly that the interactive dashboard and 3D routing diagram belong to the Claude Code build. The topology JSON survives as an inline `json` artifact with the same schema and the same controller-naming contract, now joined against the Routing Topology table instead of a 3D renderer.

### Runtime reachability — every loss fails closed

Byte-symmetry is not sufficient for a port, so each read the skill depends on was checked for reachability. None of these degrade to a clean negative:

1. **Snippet-injected route enumeration** needs `kubectl exec` into the controller pod. No shell, no pod exec, no in-cluster network path here. `traffic-routing.md` §5.5 therefore makes the blind spot a **permanent, mandatory caveat**: the report must state the true route count is **≥** the counted one, the inventory is marked Unverified, and both deep reads are handed to the cluster owner. It never infers "no hidden routes".
2. **`ValidatingWebhookConfiguration`** (`admissionregistration.k8s.io`) is likely denied → lands in the existing #153 **Unverified-means-exposed** branch, keeping the 🔴 5 CVE-2025-1974 band. The controller Deployment `args` (built-in `apps` group) is the primary signal; the VWC read is corroboration.
3. **Gateway API / CRD reads** are likely denied → Gateway API reported **unconfirmed**, with `"gatewayApi": {"readStatus": "unconfirmed"}` — never `crdsInstalled: false`, and never a "install the CRDs" recommendation on the strength of a `403`.
4. **Secret contents are never read.** TLS posture comes from `Ingress.spec.tls[].secretName` references plus the ACM inventory.

`references/porting-notes.md` carries an optional supplementary read-only ClusterRole that resolves 2 and 3 definitively.

### Scoring model carried over unchanged

Verified rather than assumed: §1.3–§1.7 of `report-generation.md` (tier mapping, deduction algorithm, category caps, gate pseudocode, bands, Score Breakdown, worked example) are **byte-identical** to upstream; Step 2 differs only by the one consistency-check row that referenced a renderer token. The §1.7 worked example still derives 10+6+6+4+4+4 = 34 → **66 HARD**, gate 1.

The **#129 learning is in** (`report-generation.md` §1.0): an empty/absent estate is not rated — 100 / TRIVIAL with a "nothing to migrate" note — including the §1.0-A carve-out that a reachable CVE/EOL controller is still a security finding at zero routes, and the orphaned-Ingress "Migration Crew Alert" path. The #153 auth-tier resolution and the #154 toolkit path with its assessor/operator fence are intact.

### Self-review before pushing (commit 2)

I ran the adversarial-review pipeline against my own port content. It caught three things worth disclosing:

- **Lens 4.** I had asserted `AmazonAIOpsAssistantPolicy` covers `networking.k8s.io`/core/`apps` but not CRDs or `admissionregistration.k8s.io`. A live check of [Review access policy permissions](https://docs.aws.amazon.com/eks/latest/userguide/access-policy-permissions.html) (2026-08-30) refutes the *sourcing*: the policy is listed among the available cluster-access policies but — unlike every other policy on that page — **its rules are not enumerated**. My group list traced to the `eks-recon` port's notes, i.e. a secondary source, and it is a negative/absence claim. Reworded at 5 sites: coverage is stated as unpublished, the group list is attributed as unverified-secondary, and the port now depends on failing closed plus the supplementary ClusterRole rather than on assumed coverage, with a `kubectl auth can-i` verification pointer.
- **Lens 10** then found that reworded claim surviving in 4 further places after my first edit — under-enumeration reproduced inside its own fix. Swept to zero, along with a surviving "score gauge" in §1.0-A.
- **Lens 12** flagged a table-column mismatch in `alb-migration.md` which I **refuted and dropped**: it is a correctly escaped `\|` inside a regex, byte-identical upstream.

### Verification (real exit codes)

All five gates EXIT=0:

| Gate | Result |
|---|---|
| `validate-frontmatter.py` | 33 skills, 0 errors |
| `update-pages.sh --check` | in sync |
| `update-all-references.sh --check` | in sync |
| `update-devops-agent-references.sh --check` | in sync |
| `devops-agent-validate` CI logic, reproduced locally | PASS all 6 ports — ours 348 KB / 33 files, no `scripts/`, no executables |

One gate violation was caught pre-commit: the copied assets came across mode 755 (upstream tracks them that way) and that CI job fails on any executable file — `chmod 644` applied.

**Blast radius: 51 files, +6470/−0, zero other files touched.** `skills/` is completely untouched. Only `devops-agent/README.md` and `misc/website/docs/devops-agent/index.md` are modified, each by exactly one generated row — no other contributor's row moved.

### Upstream observations — disclosed, not fixed

Documented in `porting-notes.md` rather than changed, since this PR does not edit `skills/`:

1. `SKILL.md` heading says **"Report Structure (5 Navigation Pages)" but the table lists 6 rows**. The port states 6 rather than carry a wrong count — a one-word upstream fix.
2. `report-generation.md:283` — the tail of the Execution-risk blockquote calls CORS / rate-limit / external auth "**no faithful equivalent**" (the literal Tier-A phrase) while §1.3 classifies those three as Tier-B *because* a workaround exists. Pre-existing, predates #129, already disclosed on #153. Carried over verbatim rather than silently diverging the port.
3. `ingress-resources.md:31` "No equivalent" vs the sibling CORS row's "No **native** equivalent". Cosmetic, disclosed on #153.

Happy to fold 1–3 into a small follow-up sweep on the Claude Code skill if you'd prefer them closed rather than disclosed.


utkarpun (changes requested):
## Review: CHANGES_REQUESTED

Thanks for a genuinely careful port — the DevOps Agent constraints are all met (33 files / 316 KB, no `scripts/`/`steering/`, no `allowed-tools`, no MCP, all mode 644), all 5 validation gates re-run clean (EXIT=0), the scoring model is verified byte-identical (10+6+6+4+4+4=34 → 66 HARD, gate 1), renderer tokens are fully de-tokenized, the #129 learning is present, and the fail-closed reachability handling is solid (Secret contents never read; 403 → `unconfirmed`, never `crdsInstalled:false`). Good work.

Blocking on a set of **technical-accuracy defects**. The key point: most of these are **byte-identical to `skills/eks-ingress-migration`** — i.e. they're pre-existing in the published skill, faithfully inherited. That's a provenance fact, not a mitigation: they now ship to DevOps Agent users unlabeled, and they're already live in the Claude Code skill. **Please fix them in BOTH copies in this PR** (`skills/eks-ingress-migration/…` and `devops-agent/eks-ingress-migration/…`), parity-preserving. No carry-over / disclose-and-accept — the operative rule is the one you already applied to the nav-count ("diverge where the value is plainly wrong"); apply it consistently here.

### Must fix (INCORRECT)

- **I-1 — fabricated annotation (highest user-harm).** `alb.ingress.kubernetes.io/certificate-discovery: 'true'` is shipped as a functional, ready-to-apply annotation (`assets/samples/alb/06-multi-host-tls-ingress.yaml:52`, `references/alb-migration.md:67`, and the byte-identical skill copies). **LBC has no `certificate-discovery` annotation** (absent at v2.7.2 → v3.5.0). Cert discovery is triggered by *omitting* `certificate-arn`; the unknown annotation is silently ignored, so an operator believes TLS discovery is active when TLS is unconfigured. Remove it and switch the sample to the omit-`certificate-arn` pattern. Ref: LBC `docs/guide/ingress/cert_discovery.md`.
- **I-2 — Gateway API production floor.** Stated as "LBC ≥ v2.14" (`references/gateway-api.md:8,20`, `SKILL.md:27`, `references/ingress-discovery.md:104`, `references/lbc-migrate-toolkit.md:18,99,100`). v2.14.0 shipped L7 Gateway API as **beta** ("restrict to testing and development"); **GA is v3.0.0** (2026-01-23). The prereq gate at `gateway-api.md:20` currently passes a beta controller as production-ready — and it self-contradicts `assets/atx/td_ingress-nginx-lbc/summaries.md:2` ("v3.0"). Correct the floor to **v3.0.0**.
- **I-3 — port-introduced self-contradiction.** `references/report-generation.md:250-254` adds a *new* verbatim mandate ("keep the literal `blocker(s) need(s)` … the wording every other surface of this skill quotes") — but the worked-example surfaces at `:144,:146,:181,:184` render the singular `1 blocker needs redesign / approval`. This mandate block is absent upstream (grep-empty), so the port introduced the contradiction. Either normalize the examples to `blocker(s) need(s)`, or scope the `:254` rule to the headline badge and drop "every other surface." (This one is port-only — skill copy unaffected.)

### Must fix (IMPRECISE-HIGH)

- **H-1 — ALB install floor vs `transforms`.** Floor stated "v2.7.2+" (`references/alb-migration.md:137,189`), but 6 of 8 ALB samples use `alb.ingress.kubernetes.io/transforms.<svc>` for URI rewrites, and `transforms` was introduced in **LBC v2.15.0** (2025-11-14) — on v2.7.2 it's silently ignored, no rewrite. Raise the floor to **v2.15.0** for the transforms rewrite path. (Both copies.)
- **H-2 — Auto Mode Gateway API overclaim.** "Gateway API / load balancing is built-in on EKS Auto Mode via the `eks.amazonaws.com` API group" (`references/gateway-api.md:10`, `SKILL.md:27`). As of 2026-08-31 Auto Mode built-in LB supports **Ingress (`eks.amazonaws.com/alb`) + Service LoadBalancer/NLB only, NOT Gateway API** — and it self-contradicts `ingress-discovery.md:129`. Scope the built-in claim to Ingress + Service; state Gateway API still needs self-managed LBC. (Both copies.)

### Must fix (narrative / disclosure)

- **M-1 / M-2 — disclosure framing.** The PR body's "Upstream observations — disclosed, not fixed (3 cosmetic items)" plus `porting-notes.md:111-118` frame the inherited content as otherwise clean, omitting I-1/I-2 (substantive, user-harming). And `SKILL.md:58`'s "faithful parity, never diverge" is presented as inviolable while you *did* diverge to fix the nav count. Since we're fixing both copies, update the disclosure to reflect that these are now fixed (not carried), and reconcile the parity-rule wording.

### Should fix (fold in while both copies are open — no carry-over)

- **S-1** `gateway-api.md:5,9,18` / `lbc-migrate-toolkit.md:33`: drop "current standard" for v1.5.0 CRDs (current is v1.6.1); keep v1.5.0 as the correct pairing for the pinned v3.4.0 tag. Do **not** touch the tag-pinned `lbc-migrate-toolkit.md:19,:99`.
- **S-2** `gateway-api.md:5` / `lbc-migrate-toolkit.md:100`: "upgrading to the current v3.4.0 release line" — current LBC is v3.5.0. Update the runtime-upgrade target (the CLI-build-tag pin at `:38-39,:99` is fine).
- **S-3** `SKILL.md:3`: negative-routes to `(eks-best-practices)`, which isn't in the DevOps Agent runtime (dead route; systemic — `eks-recon`→`eks-design` too). Strip/redirect non-subset negative-routing targets.
- **S-4 (downgraded to optional wording nit)** `ingress-discovery.md:96` / `porting-notes.md:42`: the core `apps` group "is covered" claim is *accurate* — leave it. The only mild tension is that `porting-notes.md:42` frames the policy rules as "not confirmable" alongside it; if you want, tighten that phrasing so it doesn't read as undercutting the (correct) coverage claim. Not required.
- **S-5** `SKILL.md:143` / `porting-notes.md:48`: "unlike every other access policy on that page" — `AmazonARCRegionSwitchScalingPolicy` on the same page is also unenumerated; "every other" → "most".
- **M-3**: in the narrative, distinguish *faithful* (design artifact, identity = correct) from *factual* content (versions/annotations, where identity = the error was copied).

### Advisory (non-blocking, your call)

ADV-1 `porting-notes.md` is excluded from the shipped zip (`setup.sh:162`) yet SKILL/gateway-api/ingress-discovery point at it — not load-bearing (the ClusterRole there is OPTIONAL, skill fails closed), but consider inlining the ClusterRole or rewording to "ask your operator." ADV-2 `report-generation.md:293` Tier-A phrase "no faithful equivalent" vs §1.3 Tier-B — real conflict, fix both copies eventually. ADV-3 `standard-install.yaml` only vs L4-route CRD channel (graduated in v3.5.0). ADV-5 carried-over template duplication (duplicate `## Current Configuration`, doubled Phase 3/4 sub-sections).

---

One process note: this was reviewed via our adversarial-review pipeline (12 lenses, every finding refuted before it shipped here) as the last gate. When you push the fixes, re-request review — the fix content gets its own fresh whole-branch review before merge.

kritktongrakroj:
Thanks — all six of your verifiable factual claims checked out against primary sources, and I fixed them in **both** copies as you asked. I also found three gaps in the prescription itself, one of which would have shipped a contradiction; details below.

Scope note: per your instruction this PR now edits `skills/eks-ingress-migration/` as well, so it is no longer an additive port. Two commits, deliberately split so the parity claim is auditable — `e5d480d` corrects the published skill (+ evals + its mirror), `8997a70` mirrors those corrections into the port and closes the port-only items.

## I verified your claims rather than taking them on trust

Your prescriptions are reviewable content, so each was checked against the source, not the review text. **None were refuted.**

| Claim | Verdict | Evidence |
|---|---|---|
| I-1 no `certificate-discovery` annotation | CONFIRMED | absent from the full `annotations.md` table **and** `pkg/annotations/constants.go`; `cert_discovery.md` confirms discovery fires when `certificate-arn` is *omitted* |
| I-2 production floor v3.0.0 | CONFIRMED | "not suggested for production workloads (yet!)" present at the **v2.14.0 and v2.15.0** tags, **absent at v3.0.0** (2026-01-23) |
| H-1 `transforms` needs v2.15.0 | CONFIRMED | 0 hits in v2.14.0 `annotations.md`, 5 at v2.15.0 (2025-11-14) |
| H-2 Auto Mode has no Gateway API | CONFIRMED | `auto-networking.html` + `auto-configure-alb.html` describe Service + Ingress only (2026-09-01) |
| S-1 Gateway API current v1.6.1 | CONFIRMED | v1.6.1, 2026-07-16 |
| S-2 LBC current v3.5.0 | CONFIRMED | v3.5.0, 2026-08-03 |

## Three gaps in the prescription

**1. S-1 and S-2 contradict each other as written.** You asked me to retarget the upgrade to v3.5.0 while keeping CRDs at v1.5.0. But LBC v3.5.0's own guide says it is *"built for Gateway API version v1.6.0"* and its install URL is v1.6.0. Applying both literally ships "upgrade to v3.5.0" beside "install v1.5.0 CRDs".

Resolved by stating the pairing **per controller line** — **v1.5.0 ↔ the v3.4.0 tag** (for building the CLI, which I left pinned as you said) and **v1.6.0 ↔ the current v3.5.0 runtime**. This also closes your **ADV-3** for free: Gateway API v1.6.0 moved TCPRoute/UDPRoute into the standard channel, so the experimental install is no longer needed for L4.

**2. The site lists were under-enumerated.** I swept both trees rather than editing only the cited lines:

| Item | You cited | Actually present |
|---|---|---|
| I-1 | 2 sites | 8 authored + 3 vendored, ×2 copies |
| I-2 | 7 sites | 15 ×2 copies **plus `misc/evals/…/evals.json`** |
| H-1 | 2 sites | 4 per copy |

The evals matter most — see the Lens 11 note below.

**3. M-2's citation does not resolve — pushing back.** You quote `SKILL.md:58` as presenting "faithful parity, never diverge" as inviolable. In the port, `SKILL.md:58` is `## Report Structure (6 Section Groups)`, and the strings *faithful*, *parity* and *diverge* appear nowhere in the port's `SKILL.md`:

```
$ grep -n "faithful\|parity\|diverge" devops-agent/eks-ingress-migration/SKILL.md
$ sed -n '58p' devops-agent/eks-ingress-migration/SKILL.md
## Report Structure (6 Section Groups)
```

The nearest thing is `porting-notes.md:116`, a statement about one specific carried-over item, not a general rule. So there was no inviolable parity rule to reconcile. I implemented M-1/M-3's substance anyway, since that part was sound.

## Per-item disposition

| Item | Status | Both copies (file:line) |
|---|---|---|
| **I-1** annotation removed; sample switched to omit-`certificate-arn` | DONE | `references/alb-migration.md:68`, `atx-guide.md:63,97`, `report-generation.md:448`/`:458`, `samples|assets/alb/06-multi-host-tls-ingress.yaml:52` — both trees |
| **I-2** production floor v3.0.0; reconciliation-since stated separately; prereq gate keyed on v3.0.0 | DONE | `gateway-api.md:5,8,18,19`, `SKILL.md:18`/`:27`, `ingress-discovery.md:103`/`:104`, `lbc-migrate-toolkit.md:18,33,99,100`, `migration-plan.md:22,46`, `report-generation.md:382,389,396,400`/`:392,399,406,410` |
| **I-3** badge mandate scoped to the headline badge | DONE (port-only) | `devops-agent/references/report-generation.md:254` |
| **H-1** `transforms` floor v2.15.0 at point of use + 4 floor sites | DONE | `alb-migration.md:56,138,190`, `ingress-discovery.md:103`/`:104`, `atx-guide.md:117`, `report-generation.md:460`/`:470` |
| **H-2** Auto Mode scoped to Ingress + Service | DONE | `gateway-api.md:10,19`/`:20`, `SKILL.md:18`/`:27`, `ingress-discovery.md:128`/`:129`, `lbc-migrate-toolkit.md:33`, `migration-plan.md:22`, `report-generation.md:389`/`:399` |
| **M-1/M-3** disclosure reframed as faithful-design vs faithful-defect | DONE (port) | `references/porting-notes.md:65–86` |
| **M-2** citation does not resolve | PUSHED BACK | evidence above |
| **S-1** "current" dropped from v1.5.0 | DONE | `gateway-api.md:5,9,18`, `lbc-migrate-toolkit.md:33` (tag pins at `:19,:99` untouched) |
| **S-2** upgrade target → v3.5.0 | DONE | `gateway-api.md:5`, `lbc-migrate-toolkit.md:100`, `migration-plan.md:46`, `report-generation.md:400`/`:410` |
| **S-3** dead `(eks-best-practices)` route stripped | DONE (port-only) | `devops-agent/SKILL.md:3` |
| **S-4** phrasing tension | DONE | folded into the `supplementary-rbac.md` rewrite |
| **S-5** "every other" → "most" | DONE (port-only) | `devops-agent/SKILL.md:143`, `porting-notes.md:48` |
| **ADV-1** dangling `porting-notes.md` pointer | DONE | new **shipped** `references/supplementary-rbac.md`; `setup.sh` excludes only `porting-notes.md`, so operators were being pointed at a file the packaged skill does not contain |
| **ADV-2** Tier-A phrase on the Tier-B trio | DONE | `report-generation.md:283`/`:293` — closes the residual disclosed twice on #153 |
| **ADV-3** L4 CRD channel | DONE | resolved by the v1.6.0 pairing |
| **ADV-5** template duplication | **DEFERRED** | reshapes the report template rather than correcting a fact; different regression risk, own PR |
| Nav count + `ingress-resources.md:31` | DONE | folded in, so the disclosure list is now empty |

I also **did not fork the vendored ATX bundle** (D4). `certificate-discovery` also appears in `atx/td_ingress-nginx-lbc/` — the AWS Transform definition and the AWS blog reproduced beside it, which is where the annotation originates. Editing those would silently fork them from what ATX actually ships, so they stay byte-faithful and `atx-guide.md` now carries an explicit caveat plus a review-checklist item to strip the annotation from ATX output.

## Acceptance greps (pasted output)

Redefined to the real invariant, because the caveat text necessarily contains the literal string — so a naive grep can never return zero. The invariant is **no YAML use and no prescription**:

```
$ grep -rn "certificate-discovery" skills/… devops-agent/… --include=*.yaml | grep -v /atx/
0
$ grep -rn "or \`alb.ingress.kubernetes.io/certificate-discovery\|→ \`certificate-discovery" … | grep -v /atx/
0
$ grep -rn "≥ v2.14\|>= v2.14\|< v2.14" … | grep -v /atx/ | grep -v reconciliation
0
$ grep -rn "current \*\*standard\|current standard\|current v3.4.0" … | grep -v /atx/
0
$ grep -rn "Gateway API is provided built-in\|built-in on EKS Auto Mode)\|Built-in on Auto Mode\." … | grep -v /atx/
0
$ grep -rn "no faithful equivalent" … | grep -v /atx/ | grep -v porting-notes
0
```

## What my own self-review caught in the fix content

1. **Lens 12 — I introduced a literal-asterisk site.** My ADV-2 rewording used `*precisely because*`, and `report_to_html.py`'s `inline()` has no single-asterisk em support, so it rendered as a literal `*`. Converted to bold in both copies; there is now zero single-asterisk emphasis in lines this PR adds to `report-generation.md`, the only file on the render path. The ARN glob `arn:aws:elbv2:*:*:loadbalancer/app/*` is still unmangled, which re-validates not adding a naive `\*(.+?)\*` to the tool.
2. **Lens 11 — the eval suite could not discriminate this change.** eval 4's fixture (v3.4.0) clears both the old and new floors and eval 6's (v2.12.1) fails both, so **no fixture straddled v2.14 → v3.0.0** and the floor change was untestable. Worse, **eval 6's expectation #4 asserted the old baseline and would now reject a correct answer** — the two-path bug. Rewrote eval 6 and added **eval 7** (controller v2.15.0, CRDs v1.5.0): reconciles Gateway API but sits below the production floor, which is the only case that separates "reconciles" from "production-ready".
3. A slicing bug in my own edit duplicated a `porting-notes.md` section and left the stale block behind; caught by a heading audit and repaired.

Scoring model untouched: §1.3–§1.7 are **byte-identical between the two copies** (md5 match), and the worked example still derives 10+6+6+4+4+4 = 34 → **66 HARD**, gate 1, in exactly one band.

## Verification

All five gates EXIT=0: `validate-frontmatter` (33 skills, 0 errors), `update-pages --check`, `update-all-references --check`, `update-devops-agent-references --check`, and the `devops-agent-validate` logic reproduced locally (PASS all 6 ports; ours 360 KB / 34 files, no `scripts/`, no executables, no `allowed-tools`, no MCP refs in the shipped set, no `steering/`).

Lens 12 hostile fixtures: missing input exits 1, unreadable input exits non-zero, happy path renders with this round's new content class.

**Blast radius: 46 files, all under `eks-ingress-migration`.** `devops-agent/README.md` and the website manifests did not change, so no other contributor's generated row moved.

## One pre-existing tool defect found, not fixed here

My Lens-12 hostile fixture found that `report_to_html.py` embeds **malformed topology JSON raw into the HTML** and still exits 0, printing "📊 Topology loaded" for input it failed to parse — which would put invalid JS in the dashboard. Pre-existing and not introduced by this PR; the tool is outside this round's scope. Worth its own PR alongside the known `*em*` gap, if you'd like me to take it.

Rebased onto `b754a63`. Ready for the fresh whole-branch review. Still **not** uploaded to an Agent Space, per #199.

utkarpun:
## Round-2 review — fresh whole-branch adversarial pass: **CHANGES_REQUESTED**

Thanks Krit — your round-1 work was strong: you verified our prescriptions instead of taking them on trust, caught the S-1/S-2 contradiction and the under-enumerated site lists, and were right to push back on M-2 (the `SKILL.md:58` citation doesn't resolve in the port — dropped). All five of the fact fixes I-1, I-2, H-2, S-1, S-2 re-verify clean against primary sources, the §1.3–§1.7 scoring model is byte-identical between both copies, and ADV-1 is clean.

This was a fresh **whole-branch** pass (unit = `merge-base..HEAD`, tip `8997a70`), not a delta on the fixes — so it re-audited the whole skill provenance-blind. That surfaced a batch of defects, most of which predate this PR (they trace to the original skill-add, #92) but are in scope here under the fix-both-copies ruling.

### First, correcting ourselves: H-1 floor was wrong in *both* directions

Round 1 we told you the `transforms.<svc>` floor was **v2.15.0**, and you confirmed it. **We were both wrong — the floor is v2.14.1** (independently re-verified against primary sources):

- `alb.ingress.kubernetes.io/transforms.<svc>` (the Ingress URL-rewrite annotation) shipped in **LBC v2.14.1 (2025-10-17)**, PR #4396 "ALB URL Rewrite". Present in `docs/guide/ingress/annotations.md@v2.14.1:367` and `pkg/ingress/enhanced_backend_builder.go@v2.14.1:179`; **absent** in v2.14.0 and all v2.13.x.
- The v2.15.0 value conflated it with v2.15.0's PR #4418 *HTTPRoute URLRewrite* — a Gateway API filter, a different feature. 2025-11-14 is just v2.15.0's release date.

Apologies for sending you down the wrong path there. This is **MF#1** below.

### Must-fix (10)

| # | Where (both copies + mirrors) | Issue | Fix | Source |
|---|---|---|---|---|
| 1 | alb-migration.md:56,138,190; atx-guide.md:114/117; ingress-discovery.md:103-104/153; report-generation.md ALB step 1; porting-notes.md:76; eval 7 | `transforms` floor stated as v2.15.0 | Floor = **v2.14.1** (keep "silently ignored below floor"); rebase eval 7 | LBC v2.14.1 release notes + annotations.md@v2.14.1:367 + enhanced_backend_builder.go@v2.14.1:179 |
| 2 | migration-plan.md:55-71 (Gateway example); ingress-resources.md:31-38,61-62; traffic-routing.md:53; dns-certificates.md:55 | Gateway configured via `alb.ingress.kubernetes.io/*` annotations + TLS via listener `certificateRefs` | LBC v3.x **ignores** those annotations on Gateways; docs state TLS "cannot be done via the certificateRefs field of a Gateway Listener". Use `LoadBalancerConfiguration` (scheme — default is **internal** — + listener `defaultCertificate`/`certificates`), `TargetGroupConfiguration`, `ListenerRuleConfiguration` (`authenticateOIDCConfig`). As shipped → internal ALB, no cert. HTTPRoute block itself is correct | LBC v3.5.0 docs guide/gateway/{gateway,loadbalancerconfig,targetgroupconfig,listenerruleconfig} |
| 3 | migration-plan.md:25 | `aws eks update-addon --addon-name aws-load-balancer-controller` labeled "EKS managed add-on" | No such add-on exists (live `describe-addon-versions`, 84 add-ons; EKS UG lists Helm/manifests only). Drop the add-on branch; keep the Helm path | EKS DescribeAddonVersions (live) + lbc-helm.html |
| 4 | report-generation.md:475/:465; alb-migration.md:153-154; migration-plan.md:115 | Cleanup deletes only the ingress-nginx Deployment; the `ingress-nginx-admission` ValidatingWebhookConfiguration (failurePolicy: Fail, empty selectors) survives with no endpoints → **every Ingress create/update cluster-wide fails**, including re-applying migrated ALB Ingresses and GitOps sync | Cleanup = `helm uninstall` the release (or explicitly delete the admission VWC + Service), never Deployment-only | — |
| 5 | ingress-discovery.md §1.3/§1.4; traffic-routing.md:92-98 | `tcp-services`/`udp-services` ConfigMap L4 flows invisible: nothing reads controller args/ConfigMaps/Service ports; fail-closed never fires (nothing read → nothing denied) | Add a read-only discovery step (controller args for `--tcp/udp-services-configmap` + referenced ConfigMaps + non-80/443 Service ports); extend the blind-spot caveat to L4 | — |
| 6 | report-generation.md:487-523 (ATX/Option-3 template); atx-guide.md:125-139 | The `certificate-discovery` caveat lives only in the operator workflow; the user-facing report template's ATX checklist never tells the decision-maker to strip the bogus annotation → merged ATX output leaves the listener certless | Propagate the caveat into the template's ATX Validation checklist / Option-3 steps | — |
| 7 | report-generation.md §1.7 (:170,:177) + template :246 | Worked example stipulates "NGINX 1.9.x EOL no active CVE, Impact 3", but the skill's own fail-closed rule (unverified webhook → exposed → Impact 5 + gate) puts 1.9.x in CVE-2025-1974 scope → rule-compliant score is **40 → 60/HARD, gate 2**, not 34 → 66/HARD gate 1. Template hardcodes `[[SCORE:66]][[GATE:1]]`. Teaches under-scoring a 9.8 RCE. (§1.3–§1.7 byte-identical, so fix preserves parity) | State a webhook-verified-not-exposed stipulation (+ the verification row), or move the example controller out of CVE range; sync template tokens | — |
| 8 | dns-certificates.md:35-45 §4.2 (all four copies) | "cert-manager v1.15+ has native Gateway API support"; check inspects only the pre-v1.15 `ExperimentalGatewayAPISupport` gate → a v1.15+ install *without* enablement false-passes as 🟡 Low and cert issuance fails post-cutover. Reality: beta, not default in any version; v1.15+ needs `--enable-gateway-api` (Helm `config.gatewayAPI.enabled: true`) | Check controller args/Helm config for enablement on v1.15+; unenabled → 🟠 Medium | cert-manager docs |
| 9 | assets/atx/td_ingress-nginx-lbc/transformation_definition.md:40,48,64 | Two undisclosed harmful defects in the vendored byte-faithful TD (only cert-discovery is caveated): (a) rewrite regex quadruple-escaped → decoded pattern never matches a real path (rewrites silently never fire; contradicts the TD's own blog + shipped samples); (b) proxy-read/send-timeout mapped to `deregistration_delay.timeout_seconds` (drain — wrong; correct analogue is `idle_timeout`) | Parity-preserving: extend the atx-guide.md known-defects callout to cover both (do not fork vendored bytes) | — |
| 10 | migration-plan.md:19 (skills+devops; website:30) | Phase-1 step 1 unconditionally `kubectl apply` Gateway CRDs **v1.5.0** while step 2 recommends the v3.5.0 line, which requires **v1.6.0** → operator following the file installs a mispaired CRD set | Make step 1 conditional on the controller line (v1.5.0↔v3.4.0 / v1.6.0↔v3.5.0), matching the pairing rule this PR added | LBC v3.5 docs |

MF#1 ripples into eval 7 + porting-notes.md:76 + ~8 sites/copy; MF#2 rewrites the Gateway example + 4 mapping-table rows. All fixes both copies + regenerate mirrors (`update-pages.sh`). Prescriptions above are reviewable content, not gospel — please verify the v2.14.1 floor and the Gateway-CRD mechanisms against the cited sources before landing.

### Should-fix (16, condensed)
- **S1** ALB parallel-run is unexecutable as ordered (in-place class flip drops routes); add a rename/parallel-Ingress step or state it's a low-TTL all-or-nothing cutover (Gateway path unaffected).
- **S2** §1.7 omits DNS Cutover + Downtime/Rollback rows despite "no item may be skipped" — band can flip at Impact 3.
- **S3** gateway-api.md:8 "through v2.15.x carried the warning" → persists through v2.17; write "all v2.x / through v2.17.x".
- **S4** ingress-discovery.md:103 `annotations-risk-level` didn't exist at v1.9.0 (v1.10.0+); `allow-snippet-annotations` default false only from v1.9.5.
- **S5** supplementary-rbac.md:13 `kubectl auth can-i --as-group X` errors without `--as <user>`.
- **S6** cutover tables shift DNS before `Accepted=True`; add rollback column + reorder.
- **S7** OIDC sample alb/04 references Secret `oidc-client-secret` never created; add it + "auth-* needs HTTPS listener" caveat.
- **S8** samples pin `ELBSecurityPolicy-TLS-1-2-2017-01` while alb-migration.md:71 recommends `TLS13-1-2-2021-06` — internal inconsistency.
- **S9** six rewrite pairs: nginx rewrites bare `/X`→`/` but the ALB transform regex needs the trailing slash → bare-prefix forwarded unrewritten while comments assert equivalence.
- **S10** "v2.7.2+ for the ALB Ingress path" NEEDS_CITATION — cite or drop.
- **S11** SKILL.md skip-list says "skip alb-migration.md" while export rule 3 points into it — say "loaded at export time".
- **S12** ingress2gateway (SIG-Network) absent as the NGINX→Gateway routing-translation first hop — add as a noted option (complementary, not single-hop; no LBC emitter).
- **S13** port SKILL.md:133 EC2 IAM table omits `ec2:DescribeSubnets` (policy grants it; alb-migration.md:122 needs it).
- **S14** report-generation.md:73 "deterministic — follow EXACTLY" overclaims: no count→Impact mapping for Scale/Volume.
- **S15** undated "current v3.5.0/latest v1.6.1" (×6/copy) — date once, point the rest at it.
- **S16** PR body says "33 files / 348 KB / 11 flat references" and omits supplementary-rbac.md — actual 34 files / 12 flat refs; update the body before merge.

### Needs human verification (1)
- supplementary-rbac.md:63 — `update-access-entry --kubernetes-groups` replaces rather than appends: behaviorally almost certainly true (PUT semantics) but undocumented. Soften to "observed behavior" or keep knowingly.

### Confirmed legitimate — no action
- ADV-5 (report-template duplication) deferral, and the pre-existing `report_to_html.py` malformed-JSON-exits-0 defect, are both genuinely out of scope for this round. Fine to take the tool defect in a separate PR if you'd like.

### Gate
Fix both copies + regenerate mirrors → this gets a **fresh whole-branch re-review** (not a delta on the fixes) → merge on maintainer sign-off. Not blocking your cadence otherwise.

utkarpun:
> **Round 3 of 3 — fresh whole-branch adversarial re-review** at tip `626b518a` (not a delta; the entire `merge-base..HEAD` unit was re-attacked). Prior rounds remain visible above: round 1 ([review](https://github.com/aws-samples/sample-apex-skills/pull/204#pullrequestreview-5063090600), 2026-08-31) and round 2 ([comment](https://github.com/aws-samples/sample-apex-skills/pull/204#issuecomment-5503877368), 2026-09-02).
>
> **Current branch state (2026-09-11):** the PR is now `CONFLICTING` — a later change on `main` ([the website-deps batch PR 219](https://github.com/aws-samples/sample-apex-skills/pull/219)) advanced `misc/evals/eks-ingress-migration/evals.json` after this review ran. Please **rebase onto `origin/main` first**, so the **MF-4** fix (eval 8) lands on the current `evals.json`. No other finding is affected by the rebase — I re-verified every "fix-both-copies" finding against current `main` and none has drifted.

---

# Adversarial Review (12-lens): CHANGES_REQUESTED — PR #204 round 3 (fresh whole-branch)

Reviewer pane 1 · 2026-09-09 · adversarial-review pipeline: 11 Find lenses + Lens-10 sweep + F10c gate → per-finding Verify (5 agents) → Frame → this post. 4 findings REFUTED in Verify and dropped; everything below survived refutation.

## Lenses run
| Lens | Status |
|---|---|
| 1 Structural/CI | RAN (empirical: all CI gates, 5 generators --check, hostile inputs, zip build) |
| 2 Triggering/Routing | RAN |
| 3 Eval Scaffold | RAN |
| 4 Technical Accuracy | RAN (tag-pinned gh api + live docs) |
| 5 Completeness | RAN |
| 6 Scope/Security | RAN (IAM/RBAC mapped command-by-command) |
| 7 Source Grounding | RAN (~30 claims live-checked) |
| 8 External Gap | RAN (live 2026-09-09) |
| 9 Architecture | RAN (all destructive-verb sites audited) |
| 11 Rubric Determinism | RAN (rubric re-derived by hand; byte-identity across 4 copies checked) |
| 12 Mechanical | RAN (generators zero-diff, docusaurus build clean, hostile inputs fail closed) |
| 10 Old-claim sweep | RAN (gate) — 17 old→new pairs derived; most swept clean; survivors below |

## Coverage / SHA record
Review unit = merge-base..HEAD · merge-base `652ff8e16b324749876c60221e169c844162bf85` · tip `626b518a594b95494f5e77000563e8632749211c` · diff-spec `652ff8e1..626b518a` (85 files, +7338/−208) · clone not shallow · fresh full panel at tip ⇒ accumulated coverage = whole unit. PR narrative (body + 6 commit messages) in scope and attacked (Frame). No F9 residue: all 19 agents completed.

## Port-drift gate (F10c)
gate = **COMPLETE** · main-tip `626b518a` / port-state `626b518a` (same repo/commit) · dims-checked → 1,2,3,4 · content-parity shortfalls → 2 must-fix (INCORRECT: MF-1 shared delete-deploy; MF-8 exported-to path) · constraint violations → 0 (no scripts/steering/allowed-tools/MCP; 364 KB < 6144; 33 shipped files ≤ 100) · dim-3 failures → 1 (INCORRECT: MF-12 assessor-safe CLI fence; the CloudWatch item downgraded to should-fix in Verify — fail-closed default governs) · cannot-carry (live-verified) → 1 (tools/report_to_html.py; no-scripts rule; adapted to literal-text render, verified token-by-token) · verified-ADVISORY omissions → 0 · description-parity → in-sync (ending divergence justified — no eks-best-practices port; filed IMPRECISE-low, non-blocking) · compared-units → full dim-1 unit table in the gate agent's report (available on request; every SKILL section + all 14 refs + samples + atx assets enumerated; no prior F10b artifact existed, byte-diffs run fresh).

## INCORRECT (must fix)
| # | File (canonical; fix ALL copies noted) | Claim/defect | Fix | Source |
|---|---|---|---|---|
| MF-1 | report-generation.md skills:480 / port:490 / mirrors :491,:501 — Option-2 step 6 | `kubectl delete deploy -n ingress-nginx ingress-nginx-controller` as the removal+Secrets step, unfenced (no OPERATOR marker), in the Validation column | Replace with the same file's :426-446 helm-uninstall + webhook-survivor guidance, OPERATOR-fenced | Same branch: migration-plan.md:140; report-gen :426-446 (admission webhook failurePolicy:Fail wedge) |
| MF-2 | report-generation.md :461,:478-479 (all 4 copies) — Option-2 table | Restates in-place `ingressClassName: nginx→alb` flip + "DNS weighted routing CLB→ALB" | Align with corrected alb-migration.md:163-165: renamed parallel object, verify-before-DNS; porting-notes:99's "fixed in both copies" claim is false for this table | alb-migration.md:163-165 (in-place flip = all-or-nothing, nothing to weight) |
| MF-3 | atx-guide.md:129 (all 4 copies) | "Remove NGINX Ingress Controller after validation" — bare removal, no survivor enumeration, no cross-ref | Point to the helm-uninstall/survivor guidance (webhook, IngressClass, controller Service/CLB) | Fifth/sixth cleanup surfaces uncounted by porting-notes:92 |
| MF-4 | misc/evals/eks-ingress-migration/evals.json:96,:102 — eval 8 (new) | Grades for "v2.7.2 as the correct floor for the plain ALB Ingress path" — a fact this same branch retracted as unsourceable | Rewrite expectation to the supported-release statement | alb-migration.md:152,:211; porting-notes.md:76,:99 |
| MF-5 | ingress-discovery.md skills:124 / port:125 (+ mirrors) | "Since v1.9.1 … annotations-risk-level to High. Neither is present in v1.9.0" — doubly wrong | v1.9.0 HAS allow-snippet-annotations (default true; v1.9.1 flipped it); AnnotationsRiskLevel absent in v1.9.0, default **Critical** v1.9.1–v1.11.x, **High** only from v1.12.0 | ingress-nginx config.go at tags v1.9.0/v1.9.1/v1.11.0/v1.12.0 (verified twice independently) |
| MF-6 | ingress-resources.md:70 + migration-plan.md:21 (all copies) | TLSRoute called "experimental channel CRD" | TLSRoute is **standard** channel in Gateway API v1.5.0 and v1.6.0 (the versions this skill mandates); only TCP/UDPRoute waited for v1.6.0 | standard-install.yaml release assets at v1.4.0/v1.5.0/v1.6.0 |
| MF-7 | alb-migration.md:109 (all copies) | Static CORS "via listener-rule response-header actions" | It is a listener **attribute** (`routing.http.response.*`), listener-wide ("adds the header to all responses"), not per-rule — materially narrower on grouped/shared ALBs. Also soften "WAF cannot inject CORS headers at all" (Block-action custom responses can) | docs.aws.amazon.com/elasticloadbalancing/.../header-modification.html |
| MF-8 | report-generation.md port:498 + skills:488 (+ mirrors) | Option-2 template line "Manifests exported to: `<cluster>-manifests/target/alb/`" — asserts a file write the port forbids; sibling sites already converted | Convert to "see Export Materials" like :383-385,:404,:441,:648-650 | port SKILL.md:122 (never state output saved to a path) |

## IMPRECISE-high / MISLEADING (must fix)
| # | File | Issue | Fix |
|---|---|---|---|
| MF-9 | porting-notes.md:92 (+ mirror) | "All four cleanup surfaces now say helm uninstall … OPERATOR-only" — false: six surfaces exist, two non-compliant (MF-1, MF-3) | Correct the count and the claim after fixing MF-1/MF-3 |
| MF-10 | alb-migration.md:115,:143; atx-guide.md:124; SKILL.md:246 (all copies) | EKS Auto Mode offered as drop-in ALB-path/transforms substitute. Auto Mode docs (2026-09-09): `auth-type: oidc` **"Not supported"**, `group.name` IngressClass-only, `transforms` unmentioned (and ListenerAttributes "Not yet supported" cuts against) — the Basic-Auth→OIDC tier and ALB grouping silently no-op on Auto Mode | Add Auto Mode annotation-subset caveats at each site; drop or source the transforms-floor equivalence at atx-guide:124 | auto-configure-alb.html Considerations |
| MF-11 | ingress-discovery.md §1.4 + whole tree (all copies) | The skill never states the retirement HAPPENED: kubernetes/ingress-nginx archived ~2026-03-23, final releases 2026-03-19, every version permanently unpatched. §1.4 tells the assessor to consult a "supported/EOL matrix" that no longer exists; a healthy v1.15.1 estate scores Currency clean and the report frames migration as optional (Frame: healthy estate reads TRIVIAL ⇒ "no need to migrate") | State the archival as fact (past tense, dated) in §1.4 + report Overview/Exec-Summary baseline | gh api repos/kubernetes/ingress-nginx (archived:true) |
| MF-12 | port lbc-migrate-toolkit.md:5,:35-44,:71,:75 (+ porting-notes:58,:63) | "Safe for the **assessor** to run": lbc-migrate CLI, git clone/make, `--console` web UI, kubectl jsonpath — in a runtime with no shell/CLI/file writes (port SKILL.md:108,:120-122). Moderated: file loads at report-authoring time, but the fence affirmatively invites the agent to run tools it lacks | Re-cut the fence for the port: all lbc-migrate steps are owner/local-machine actions (traffic-routing.md:98 pattern); reachable equivalent for :75 = raw-API GET of the dry-run-plan annotation |
| MF-13 | atx-guide.md:127-129 (all copies) | Steps: TD flips `ingressClassName` in place (TD step 2, verified) → then "DNS weighted routing to shift traffic" — nothing to weight once the flip applies; contradicts alb-migration.md:163-165 doctrine. TD is byte-faithful (leave it); the guide's author-owned steps must carry the all-or-nothing/parallel-object warning |
| MF-14 | PR body (live, verified verbatim today) | "Blast radius: 51 files, +6470/−0, zero other files touched", "`skills/` is completely untouched", "this PR does not edit `skills/`", "33 files, 348 KB" — tip reality: 85 files +7338/−208, 20 `skills/` files modified, evals.json edited, port 34 files/364 KB; claims are unscoped present-tense and R1/R2 commits are unmentioned. Legit fix-both-copies work, but the body tells a merger the published skill is untouched | Update body to tip numbers + "edits both copies"; also "+6470" ≠ 994dca8's +6462 |
| MF-15 | gateway-api.md:11 (3 copies; NEW this branch) | "Checked 2026-09-09: latest upstream Gateway API release = v1.6.1 (2026-07-16)" — v1.6.2 published 2026-09-03; the self-declared currency single-source-of-truth is wrong at its own check date (CRD pairing v1.6.0 unaffected) | Update to v1.6.2 + re-date |
| MF-16 | migration-risk.md:43 (all 4 copies) | "nginx basic auth → ALB + Cognito/OIDC (**Gateway annotation**)" — old mechanism; annotations are silently ignored on a Gateway; corrected elsewhere to the ListenerRuleConfiguration CRD via parametersRef | Name the CRD mechanism |

## NUANCED / should fix
| # | File | Issue | Both sides |
|---|---|---|---|
| SF-1 | ingress-resources.md:44 (all copies) | Low-impact criterion "map to HTTPRoute features or **Gateway annotations**" vs same file's :41 "annotations do not work on a Gateway" | Real gateway.k8s.aws/* annotations exist (dry-run), but within this doc's framework the targets are HTTPRoute features + CRDs — reword |
| SF-2 | port iam-policy.json + SKILL.md:130-137 | No `cloudwatch:GetMetricData/GetMetricStatistics`, yet the verify-before-downgrade checks read LB metrics. Downgraded from must-fix in Verify: fail-closed "treat as live" governs, so degradation is designed, not broken — but the "ready-to-use" policy silently forces worst-case scoring on every orphan estate | Add the two read actions, or add a CW-named degraded-read disclosure |
| SF-3 | report-generation.md:124 vs :61,:212 + SKILL summaries (both copies) | Re-arch gate counts "production routes" in exactly one of four statements; "production" undefined. Verified: eval 2's outcome unaffected (worked example counts unconditionally), but real estates can fork | Drop the qualifier or define it |
| SF-4 | atx-guide.md:10,:41-55,:143 (all copies) | ATX workflow (SA onboarding, workspace upload, Transformation Job) lags current AWS Transform custom-TD docs (skill-format SKILL.md TDs, `atx custom def` CLI, publishable-dir rule); vendored bundle shape can't be published via current CLI without reshaping | Byte-faithful intent (atx-guide:70) explains the bundle; the guide's own operational steps still need updating to the documented CLI flow |
| SF-5 | port report-generation.md:288 (+ mirror :299) | "wrap the most damaging facts in `** **` so they render red" — bold, not red; residual of upstream's removed `!! !!` token; contradicts the port's own rule 10 (:236) | Drop "render red" |
| SF-6 | report-generation.md:5 (all copies) | "5-section navigation structure" vs the operative six-group map (:616-621) + both SKILL.mds. Verified IMPRECISE-low (stale summary; render governed by the map) | Update count + add References to the list |
| SF-7 | atx-guide.md:74-76 (all copies) | Says the TD maps proxy timeouts to `deregistration_delay` — TD:64 actually offers "…or ALB idle timeout settings" (the correct analogue) as an alternative | Quote the TD accurately; keep the manual-review ask |
| SF-8 | alb-migration.md annotation map | Verified truly unmapped tree-wide: `backend-protocol-version: GRPC`, `permanent-redirect`, `app-root`, `custom-http-errors` (affinity + canary claims were REFUTED — covered elsewhere) | Add rows or documented-removal notes |
| SF-9 | Survivor-gap set (all verified): (a) non-Helm teardown never enumerates the controller's main LoadBalancer Service → orphan CLB/SGs/DNS (migration-plan:140, alb-migration:175); (b) external-dns stale-record cleanup absent from dns-certificates.md; (c) TD step 10 removes TLS Secrets with no cert-manager re-issue caveat; (d) migration-plan Phase 4 omits the IngressClass removal report-gen has; (e) lbc-migrate-toolkit:81 lacks the shared-IngressGroup ALB caveat | Enumerate survivors + detection per site |
| SF-10 | ingress-discovery.md:31 (all copies) | "nginx-ingress" (F5's name) listed among discoverable controllers while all guidance is community-`ingress-nginx`-specific; no disambiguation | Add one community-vs-F5 line (nginx.org/* annotations out of scope) |
| SF-11 | report-generation §1.3-1.5 (all copies) | Score presented as "deterministic/auditable" with no disclosure that weights/caps/bands are hand-tuned heuristics; customer-facing headline reads calibrated. Mitigated by footer disclaimer + "not a manday estimate" | Add one sentence: heuristic weighting, not an empirically validated model |

## Needs Human Verification
None — every surviving finding is CONFIRMED. (One open question folded into MF-10: whether Auto Mode supports `transforms` is undocumented either way as of 2026-09-09 — the defect filed is the unsourced positive claim, not the capability.)

## Unverified residue / uncovered scope
None. All 11 lenses + both gates + 5 verifiers + Frame completed; 4 findings refuted and dropped (LBC-floor phrasing, Tier-B escalation "fork" — reconcilable via the Tier-B cap, the gateway-api.md "never loaded" claim — SKILL:77 is a live loader, and the external-dns overwrite mechanism — TXT-registry ownership protects manual records).

## Advisory (post-merge, non-blocking)
- external-dns: no duplicate-source sequencing/exclusion guidance (the real risks are owned-record flap + R53 mixed-routing-policy rejection, NOT overwrite of manual weighted records — verified) — worth one paragraph in dns-certificates.md.
- devops-agent/eks-recon routes "before a migration" queries but never points to the new port; port exclusions omit eks-security (sibling file untouched by this PR — file separately if desired).
- iam-policy.json `ec2:DescribeSecurityGroups` granted but no doc consumes it; elbv2 `Describe*` wildcard could be enumerated.
- Samples use `rewrite.bar.com` (real domain; upstream-copied) — RFC-2606 name preferred.
- No rollback abort-thresholds/bake-time/TTL-prelower steps (cutover legs say "monitor and shift back").
- Undated negative claims ("no LBC EKS add-on", "no AWS emitter") verified true 2026-09-09 but carry no as-of anchor.
- 16 skills-copy sample YAMLs are mode 100755 (pre-existing; port copies correctly 100644) — chmod follow-up.
- Template duplication (dup "## Current Configuration", doubled Phase blocks) disclosed as deliberately deferred (porting-notes:102) — tracked, not re-flagged.
- Zero eval fixtures mention the retirement (ripple of MF-11); 340-line vendored AWS blog republished as website pages; ATX machine-input files render as sidebar pages.
- Pre-existing, NOT this PR (for the repo defect lane): update-devops-agent-references.sh silently drops a row on missing SKILL.md (exit 0); eks-operation-review porting-notes LICENSE dead link; terraform-skill anchor warning; website evals-manifest regen drift.

## Architecture concern (prose)
~150 version literals replicated across 4 trees with a single currency-note guard that at least one file still half-bypasses; the vendored-blog republication will read staler every month. The port's do-vs-instruct split is otherwise architecturally clean (OPERATOR fences held up everywhere except MF-12).

## Strengths (coordinator synthesis)
- All wired CI gates pass empirically; every generator regenerates byte-exact; hostile inputs fail closed; docusaurus builds with zero PR-attributable broken links.
- The scoring rubric re-derives correctly (caps sum 125, worked examples exact, bands complete/disjoint) and is byte-identical across all four copies.
- F10c dim-2 is fully clean (no scripts/MCP/allowed-tools; 364 KB / 33 shipped files) and the dim-1 unit table shows genuinely thoughtful adaptations (HARD-STOP multi-cluster, inline topology contract, fail-closed degraded reads).
- Carried facts held: v2.14.1 transforms floor and CRD-driven v3.x Gateway config are correct everywhere, including the new evals.
- porting-notes' self-critique is honest (Frame found its parity narrative clean apart from the "four surfaces" count).

## Verdict
**CHANGES_REQUESTED** — 16 must-fix (8 INCORRECT, 8 MISLEADING/IMPRECISE-high), 11 should-fix, advisories above. Gate for next round unchanged: fresh whole-branch re-review of the fixes, merge only on user sign-off. Fork PR, append-only, contributor squash.

kritktongrakroj:
Thanks — this round found the thing I'd have most regretted shipping. **MF-11 is the one that mattered**: the skill described an EOL/CVE band in detail and never once said the retirement had actually *happened*, so a healthy v1.15.1 estate scored Controller Currency clean and the report read as "migration optional". That's a wrong conclusion from correct parts, and it survived three rounds of review including my own sweeps.

All 16 must-fix and 11 should-fix are addressed in both copies. I verified your prescriptions against primary sources first; **two did not survive unchanged**, and I've scoped the wording so the guidance no longer depends on the part I couldn't confirm.

Rebased onto `237e2ea` first, as you asked. Two commits, split so the parity claim stays auditable — `eb50e17` corrects the published skill + evals, `d8230d2` mirrors into the port and closes the port-only items.

## The rebase conflict resolved itself in your favour

`evals.json` was the only conflict. #219 **stripped `expected_output` from every eval**, so I accepted that convention and dropped it from our evals 7 and 8 too — which deleted **half of MF-4's surface** outright. The offending "v2.7.2 as the correct floor" sentence lived in both `expected_output` (now gone) and `expectations` (rewritten to grade the real invariant: the floor correction is scoped to the rewrite path).

## I verified before implementing

| Prescription | Verdict | Evidence |
|---|---|---|
| **MF-11** ingress-nginx archived, permanently unpatched | **CONFIRMED — date corrected** | Repo banner: *"This repository was archived by the owner on **Mar 24, 2026**. It is now read-only."* — **2026-03-24**, not ~03-23. README retirement notice: after March 2026 there are *"no further releases, no bugfixes, and **no updates to resolve any security vulnerabilities**"*. Supported-versions table frozen at **v1.15.1**. |
| **MF-15** currency anchor stale at its own check date | **CONFIRMED** | Gateway API **v1.6.2 released 2026-09-03**. `gateway-api.md:11` said "checked 2026-09-09: latest = v1.6.1". Embarrassing for a note whose whole job is currency. |
| **MF-6** TLSRoute is not experimental-channel | **CONFIRMED** | v1.5.0 release notes: *"TLSRoute and ListenerSet have graduated to the **Standard channel as `v1`**"*. Both CRD versions this skill pins carry it. Only TCPRoute/UDPRoute waited for v1.6.0. |
| **MF-5** v1.9.0 snippet claim doubly wrong | **CONFIRMED (directly)** | `config.go@controller-v1.9.0`: `AllowSnippetAnnotations` **exists** and `NewDefault()` sets it **`true`**. So v1.9.1 is where the default *flipped*, not where the setting appeared. |
| **MF-7** ALB CORS is not a per-rule action | **CONFIRMED** | ALB docs: response headers are **listener attributes** and the LB *"adds the header with the configured value to **all responses**"*. |
| **MF-10** Auto Mode is not annotation-equivalent | **CONFIRMED, and worse than filed** | Auto Mode annotation table lists `auth-type: oidc` ("OIDC Auth Type is currently not supported"), `group.name`, `waf-acl-id`/`web-acl-id` and `dry-run-plan` as **Not supported** — *and* **"You cannot set `ListenerAttribute` with EKS Auto Mode."** |

**MF-10 and MF-7 interact, which neither finding noticed.** The static-CORS workaround *is* a listener attribute — so on Auto Mode it isn't merely narrower, it's **unavailable entirely**, and the WAF fallbacks the sibling Tier-B features lean on are constrained too (`waf-acl-id`/`web-acl-id` unsupported, `wafv2-acl-name` "Upcoming soon"). Both copies now say so.

## ⚠️ MF-11: I took option (b) — a scoring floor — and the published example does not move

A floor is now in §1.4: **a *running* ingress-nginx controller scores minimum 🟠 3** on Controller Currency, because the estate carries a standing, unfixable security-patch gap. Documentation alone would have left the defect alive — a clean Currency row is exactly what told the reader migration was optional.

It is deliberately conservative:

- The floor only ever **raises** a band. The 🔴 5 conditions still govern.
- **🟢 0 still applies** to an absent or fully-down controller — nothing running, nothing exposed.
- It is scoped **to ingress-nginx specifically**; other controllers are still judged against their own live support matrices.

**Arithmetic check: nothing moved.** The §1.7 worked example's controller row was already **Impact 4**, above the floor, so `10+6+6+4+2+1+4+4+0 = 37 → 63/HARD, gate 1` stands and `[[SCORE:63:HARD]] [[GATE:1]]` is untouched. The 57/VERY HARD counterfactual is likewise unaffected (its controller is 5).

**One Lens-11 consequence I want on the record:** the floor makes the **🟡 1–2 band unreachable for ingress-nginx**. It is **not vacuous** — it stays reachable for any other EOL controller — but that is a direct result of scoping the floor to ingress-nginx, and it's the kind of thing worth re-checking if the scope ever widens.

## Two prescriptions I did not adopt as written

**1. The `annotations-risk-level` default history.** You said `Critical` for v1.9.1–v1.11.x and `High` only from v1.12.0. I confirmed the field is **absent at v1.9.0** and that the **current** default is `High`, but I could not verify that Critical→High transition per-tag in this environment. Rather than assert a version boundary I hadn't checked, I wrote the guidance to **not depend on it** — it states the verified facts and adds the operationally important one: default `High` **admits** High-risk annotations and blocks only `Critical`, so a default-configured controller is *not* "hardened". If you have the per-tag evidence, I'll fold the boundary in.

**2. "Final releases 2026-03-19."** Not cited, because I didn't verify it. The archival date and the no-more-security-patches statement are both sourced; the release date isn't, so it isn't in the text.

## Round 2's own errors, corrected

Two of these were mine from last round, and the fix content is where the review process keeps finding them:

- The **v1.6.1** currency note was already superseded when I wrote it → now **v1.6.2** (2026-09-03).
- I claimed **"all four cleanup surfaces"** were fixed. There are **six** — the Option-2 steps table and `atx-guide.md`'s post-transform steps were never touched, and those are exactly the two MF-1 and MF-3 caught. `porting-notes.md` now records the correct count and that round 2 overstated it.

## Per-item disposition

| Item | Status | Where (both copies unless noted) |
|---|---|---|
| **MF-1** unfenced `kubectl delete deploy` in a Validation column | DONE | `report-generation.md` Option-2 steps → step 7 `helm uninstall`, **OPERATOR**-marked |
| **MF-2** in-place flip + "DNS weighted routing" | DONE | Same table: renamed parallel object, verify-before-DNS, all-or-nothing fallback named |
| **MF-3** bare "Remove NGINX Ingress Controller" | DONE | `atx-guide.md` steps 5–8, with survivors enumerated |
| **MF-4** eval 8 graded a retracted claim | DONE | `evals.json` — `expected_output` stripped per #219, expectation rewritten |
| **MF-5** v1.9.0 snippet-hardening | DONE | `ingress-discovery.md` — `allow-snippet-annotations` exists at v1.9.0 (default `true`) |
| **MF-6** TLSRoute channel | DONE | `ingress-resources.md`, `migration-plan.md` pairing table |
| **MF-7** CORS listener attribute + WAF absolute | DONE | `alb-migration.md` fidelity-gap note |
| **MF-8** file-write path in the port | DONE | Both copies now defer to **Export Materials** (the skills path also contradicted its own Step 6) |
| **MF-9** "all four cleanup surfaces" | DONE (port) | `porting-notes.md` — six, and round 2 overstated |
| **MF-10** Auto Mode drop-in | DONE | `alb-migration.md` ×2, `atx-guide.md`, `SKILL.md` Auto Mode step |
| **MF-11** retirement + Currency floor | DONE (option b) | `ingress-discovery.md` §1.4, `report-generation.md` rule 15 |
| **MF-12** port fence invites unrunnable tools | DONE (port) | `lbc-migrate-toolkit.md` — all steps **OWNER**; console → read-only API GET of `dry-run-plan` |
| **MF-13** ATX in-place flip + weighting | DONE | `atx-guide.md` step 5 |
| **MF-14** stale PR body | **owner action** | Replacement body drafted — see below |
| **MF-15** currency anchor | DONE | `gateway-api.md` → v1.6.2, re-dated 2026-09-13 |
| **MF-16** "Gateway annotation" for auth | DONE | `migration-risk.md` → `ListenerRuleConfiguration` + `parametersRef` |
| **SF-1** low-impact criterion self-contradiction | DONE | `ingress-resources.md:44` |
| **SF-2** CloudWatch reads | DONE (port) | `iam-policy.json` + SKILL.md IAM row explaining the fail-closed degradation |
| **SF-3** undefined "production routes" | DONE | Qualifier dropped from the one gate statement carrying it |
| **SF-4** ATX workflow lags current docs | DONE | `atx-guide.md` — verify the current custom-definition flow; bundle is reference content |
| **SF-5** "so they render red" | DONE (port) | Port emits `**bold**`; colour belongs to the upstream renderer |
| **SF-6** stale "5-section" nav | DONE | → six, References added |
| **SF-7** ATX timeout quote | DONE | TD does offer idle timeout — as the *second* option |
| **SF-8** unmapped annotations | DONE | `backend-protocol-version: GRPC`, `permanent-redirect`, `app-root`, `custom-http-errors` |
| **SF-9** survivor gaps (a–e) | DONE | Both Phase-4 blocks, external-dns TXT cleanup + source sequencing, cert-manager re-issue warning, IngressClass + controller `LoadBalancer` Service, shared-IngressGroup warning |
| **SF-10** F5 vs community nginx | DONE | `ingress-discovery.md` discovery list |
| **SF-11** score presented as deterministic | DONE | §1.3 — hand-tuned heuristic, not a validated model; the *arithmetic* is what's deterministic |

## Verification

**All five gates EXIT=0** — `validate-frontmatter` (33 skills, 0 errors), `update-pages --check`, `update-all-references --check`, `update-devops-agent-references --check`, and `devops-agent-validate` **reproduced verbatim from the workflow file** → PASS all 6 ports (ours **420 KB / 34 files, 33 shipped**).

> Worth flagging as a process note: my first hand-rolled version of that last gate added an `allowed-tools` grep **the real CI does not perform**, and it "failed" two ports — including one this PR never touches. Running the workflow's own script instead of a paraphrase is the difference between a real gate and a false alarm. (For the record the only `allowed-tools` string in the port is a *porting-notes* table row documenting that the constraint exists — and porting-notes isn't even shipped.)

- **Lens 10** — swept 15 changed facts across all four copies + evals: **zero true survivors.** Five hits adjudicated as legitimate: `non-production cluster` (different word), the skills copy's real export path (port = 0), `!! !!` which `report_to_html.py:34` genuinely renders red, a `v2.7.2 → v3.5.0` version *range*, and `assessor-safe` — which is **correct** in the shell-capable Claude Code runtime, since MF-12 is a port-only defect. The last two are intentional copy asymmetries, not misses.
- **Lens 11** — worked example re-derived by hand (37 → 63/HARD gate 1, unchanged); §1.3–§1.7 **byte-identical across both copies** (md5 match), including the new SF-11 disclosure.
- **Lens 12** — missing input **exit 1**; happy path on a fixture carrying this round's new content classes (floor blockquote, OPERATOR row, parallel-object step, the new annotation rows, the ARN glob) renders clean: score + gate badge present, 6 `<strong>`, 1 hot span, **zero odd table-cell counts**, ARN glob unmangled.
- **Phase 3** — 21 parity items confirmed **Y/Y** by independent grep in each copy, 3 port-only items **N/Y** as designed. Two initial misses were my own regex escaping, not missing edits.

**Blast radius: 48 files this round, all under `eks-ingress-migration`.** `devops-agent/README.md` and the website manifests did not move, so no other contributor's generated row changed.

## Still deliberately out of scope

**ADV-5** (report-template duplication) and the pre-existing `report_to_html.py` malformed-JSON-exits-0 defect, as agreed. I re-confirmed the latter is live *and* that `git diff` shows **the tool is untouched by this branch** — so it's provably inherited, not introduced. Happy to take it with the `*em*` gap in its own PR.

utkarpun (approved):
Round-4 fresh whole-branch adversarial re-review (12-lens) at tip `d8230d2e` — APPROVED.

All round-3 findings verified genuinely fixed; no INCORRECT and no high-severity imprecision survive. Port-drift gate COMPLETE (432 KB / 34 files, read-only, fully autonomous, constraints clean, registry in sync); all CI generator + eval-hygiene gates green; the load-bearing version facts (transforms floor v2.14.1, Gateway API v3.0.0 GA, TLSRoute Standard v1.5.0, TCP/UDPRoute Standard v1.6.0, snippet-hardening + annotations-risk-level v1.9.1, CVE-2025-1974 fixes, EKS Auto Mode 'Not supported' matrix) were re-confirmed against tag-pinned primary sources; the scoring rubric re-derives deterministic (37 -> 63 / HARD / gate 1); teardown-survivor enumeration is complete with per-survivor detection.

A small set of low/medium doc nits (a dead `assets/atx` path reference, a supplementary-RBAC 'grants exactly the reads' wording overclaim, a 'mirror -> no Gateway API equivalent' imprecision, a missing cert-manager caveat in the ATX transform spec, and two citation-precision items) are non-blocking and will be handled as post-merge follow-ups. Merging via squash.

Issue #14, opened 2026-03-26
Improve Upgrades Workflow and Skill


utkarpun:
List out issues:  
- Currently on check if CLI or Terraform used to deploy. Need to account for IaC and CI/CD more thoroughly. Should just ask user instead of assuming.  
- Do a proper split between In Place and Blue Green. With Blue Green properly leveraging our EKS BG Repo.
- Instead of cloning repo, just have it there to avoid sed errors
- Also integrate the steering file better with both EKS MCP Server and EKS Reconnaissance tool

utkarpun:
Related context: the `steering-workflow-creator` skill (feat branch `feat/steering-workflow-creator`, at `skills/steering-workflow-creator/`) codifies a strict steering-workflow convention. Running its linter against `steering/workflows/upgrade.md` surfaces 48 errors + 1 warning: missing `## Defaults`, `## Quality Checklist`, `## Conversation Style` sections; phase headings use flat `## Phase N:` instead of `## Phases` + `### Phase N:`; 16 `--`/`->` drifts (should be `—` / `→`); file length 442 lines, close to the 450 cap.

The items already listed in this issue (IaC/CI-CD detection, In-Place vs Blue-Green split, avoiding repo cloning, MCP/eks-recon integration) can all land under the convention. Proposed scope additions:
1. Bring `upgrade.md` into convention compliance — run `python skills/steering-workflow-creator/scripts/quick_validate.py steering/workflows/upgrade.md` until it exits 0.
2. Per `references/tool-routing.md`, annotate each phase with `Source:` and name the CLI fallback for any `live` phase.
3. Wire `eks-recon` as the Phase 1 discovery skill and `eks-mcp-server` as the setup-bridge where `live` phases need cluster state.

This retrofit was scoped OUT of the skill-creator PR deliberately so the convention ships first.

utkarpun:
Superseded by eks-upgrade-check (PR #36). Old upgrader lives on experimental/eks-upgrader (unmaintained).

PR #210, opened 2026-09-04
feat(eks-war-skill): Add EKS Well Architected Review Skill

terjing:
## Summary

Add EKS Well Architected Review Skill

- [x] Ran `/apex:new-skill` (or walked the equivalent manual steps in `CONTRIBUTING.md`)
- [x] `skills/<skill>/SKILL.md` present with valid frontmatter
- [x] `misc/evals/<skill>/triggering.json` authored (≥16 prompts; balanced positives and near-miss negatives)
- [x] For each neighbour: `misc/evals/<neighbour>/SIBLING_MAP` gained a bullet and its `triggering.json` gained the matching negatives (via `update_sibling_map.py` or hand-edit)
        - ~currently have no neighbours~ added negative sibling map
- [x] Ran the `update-docs` skill and committed any resulting changes (regenerated wrappers/manifest, marker-block updates, prose edits)


utkarpun (changes requested):
# Review: CHANGES_REQUESTED (whole-branch, tip `339371a`)

Thanks Ter Jing. This skill has real strengths (noted at the end), and this is a fresh full review of the whole branch as it landed here, not a delta against the earlier version we looked at. Every finding below was re-verified before posting: against live AWS documentation as of 2026-09-04, or by executing the shipped scripts/jq on constructed fixtures. Review unit: `21ef66db..339371ac` (40 files, +10,349/-4). CI passes; none of the must-fix items are CI-caught.

**Verdict: 20 must-fix (15 incorrect, 4 misleading, 1 high-impact imprecise) and 9 should-fix.** Two label assumptions that AWS docs are silent on were settled by running the check on a live EKS 1.34 cluster with an active Fargate profile: one came back in the skill's favor, one is a confirmed bug (finding 29 below; details in the Fargate section).

---

## Must fix: incorrect

| # | Where | Problem | Suggested fix |
|---|---|---|---|
| 1 | `SKILL.md:218,227-229` (+ website twin) | Claims "lists are checked against the scorer's own N/M and gate 11 fails on any disagreement", but no such gate exists in the branch. `render-report.py:1477-1478` (`else: note = ""`) silently renders a contradicting list and exits 0. Reproduced by execution, twice independently. | Implement the gate (visible disagreement marker plus non-zero exit or explicit caveat), or delete the claim. |
| 2 | `references/cost-optimization.md:55` | cost-8 detail emits failures/total ("N/M unattached") while the renderer's extractor and every sibling detection use passes/total. Guaranteed false disagreement whenever idle != total-idle, and a false agreement at the exact midpoint. Verified by execution. | Emit passes/total like the siblings (or invert the extractor). |
| 3 | `SKILL.md:229,236-238` | "Currently 48 extractors, 585 agreements, 0 disagreements across the fixture suite" plus 0.33ms/0.06ms benchmarks: the shipped registry has 65 entries (AST-counted), and no fixture suite ships anywhere in the branch. The numbers are not reproducible from the tree. | Delete, or replace with facts reproducible from the tree. |
| 4 | `references/performance-efficiency.md:28-36` (+ `render-report.py:778-780` twin) | perf-3's regex classifies m5/c5/r5/t2/i3/i3en/d2/h1/x1/x1e as previous-generation. AWS's actual previous-generation list is A1, C1, C3, C4, G3, I2, M1-M4, P3, P3dn, R3, R4, T1: 10 of the comment's "eleven real families" are current generation. An all-m5 fleet renders a factually false "0/N current-generation". Source: EC2 instance types docs, previous-generation and current-generation pages, checked 2026-09-04. | Re-pin the regex to AWS's actual list and add a guard to re-derive it live. |
| 5 | `references/*.md` (~35 lines, e.g. `operational-excellence.md:79`) vs `SKILL.md:333` and the reducer | Interactive governance answers: the references tell the user to emit `not-applicable`, but SKILL.md and the reducer only special-case `na`. "Doesn't apply" is therefore scored as an answered 0. Executed: 2x`all` + 3x`not-applicable` scores 40 instead of 100. | Standardize on `na`, or normalize the token in the reducer. |
| 6 | `references/security/identity-access.md:89` (+ `render-report.py:347` twin) | sec-30 ("no ssh open") false-passes a world-open security group: `IpProtocol: "-1"` rules carry no FromPort/ToPort (the EC2 API marks ports optional; -1 means all ports), so `(.ToPort//0)>=22` is false. Reproduced on both rule variants. | Add an `IpProtocol == "-1"` clause to both copies. |
| 7 | `references/security/data-protection.md:74` | sec-21 remediation `aws ec2 modify-volume --volume-id <id> --encrypted` is impossible: ModifyVolume has no encryption parameter and EBS cannot encrypt in place. The documented path is snapshot, encrypted copy, restore (or account-level default encryption for new volumes). | Replace with the documented path. |
| 8 | `references/security/identity-access.md:159` | sec-2 remediation says to restrict public API endpoint access via the cluster security group. The actual control is `publicAccessCidrs` via `aws eks update-cluster-config`; security groups cannot restrict the public endpoint. | Point at `update-cluster-config` / `publicAccessCidrs`. |
| 9 | `references/security/network.md:110-122` + scorer in `identity-access.md:94` | net-4 awards `all` for "separate SGs" when `securityGroupIds` lacks `clusterSecurityGroupId`, but EKS by design applies the cluster SG to the control plane and managed nodes. The measured premise contradicts the documented default and the recommended architecture. | Reframe the question and scorer (for example node-SG hardening), or drop it. |
| 10 | `SKILL.md` Steps 1-4 + `references/workflow.md` (gap) | No cluster-identity binding: no `--context`, no kubeconfig-endpoint vs `.cluster.endpoint` cross-check, no confirm step. The validation gate passes with the AWS half from cluster A and the kubectl half from cluster B. Sibling precedent exists: eks-recon carries exactly this guard (`SKILL.md:211`) and eks-operation-review pre-flights `update-kubeconfig`. | Add a context/endpoint cross-check before collection. |
| 11 | `SKILL.md:101-111` + all scorers (gap) | A functionally dead cluster bands normally: the viability gate is `nodes==0 AND workload_pods==0`, and there are zero node-Ready / pod-phase / replica-readiness / addon-health reads anywhere (grep-verified; only PV/PVC phases). NotReady nodes keep labels and nodeInfo, so spec-side jq passes; 0 nodes with Pending pods bypasses the AND. This undercuts the description's own promise that under-observed clusters cannot score well. | Gate on node Ready count and pod phase, and add a liveness caveat next to the band. |
| 12 | `SKILL.md:4` | `license: Apache-2.0` violates CONTRIBUTING.md:458-460 (in-house skills must not declare a `license:` field; the repo LICENSE is MIT-0). There is no LICENSE / UPSTREAM.md / .vendor-sha / THIRD_PARTY_NOTICES set, and none should be added: there is no genuine third-party upstream here, so adding provenance files would misattribute the work. | Drop the `license:` key. |
| 13 | PR checklist + `misc/evals/eks-well-architected-review/README.md:13` | "Currently have no neighbours" is not accurate: the skill's own description names five sibling exclusions, and the PR itself edits a neighbour. Repo convention makes sibling eval fan-out mandatory when neighbours exist (new-skill.md:82,93; CONTRIBUTING.md:324; precedents `17e7466` fanning out to 4 skills and `2a2d33f` to 5). No sibling negatives or SIBLING_MAP bullets shipped, and the closest triggering collision (eks-operation-review: "any request to audit, review ... or score") is untested on both sides. | Correct the checklist and ship the repo-owned fan-out: near-miss negatives plus SIBLING_MAP bullets under the neighbours' `misc/evals/` (op-review, cost-intelligence, recon, best-practices, security). Note: do not edit the vendored op-review/upgrade-check SKILL.md descriptions; those are upstream-owned. |
| 14 | `references/operational-excellence.md:55,514,529` + collection list in `references/workflow.md` | fargate-4 credits only a `fluent` sidecar, but EKS Fargate's documented mechanism is the built-in log router configured via the `aws-logging` ConfigMap in the `aws-observability` namespace (the docs are explicit that you do not run Fluent Bit as a sidecar). ConfigMaps are never collected, so a compliant cluster can never be credited, and the remediation prescribes the outdated sidecar pattern. | Collect and check the `aws-observability/aws-logging` ConfigMap; fix the remediation. |
| 29 | `references/operational-excellence.md:53,55` (fargate-2/fargate-4 pod selection) | Fargate **pods** do not carry `eks.amazonaws.com/compute-type`; their auto-applied label is `eks.amazonaws.com/fargate-profile: <profile>`. Verified on a live EKS 1.34 Fargate cluster (evidence below). Both detections therefore match zero pods on a real Fargate cluster and silently no-op to `na`. Numbered 29 to keep the IDs below stable. | Select pods by presence of the `eks.amazonaws.com/fargate-profile` label key (any value), or resolve via the pod's `nodeName` to a node with `compute-type=fargate`. Interacts with finding 14: once selection works, fargate-4 still needs the ConfigMap check. |

## Must fix: misleading framing

| # | Where | Problem | Suggested fix |
|---|---|---|---|
| 15 | `SKILL.md:3` + skills/README + manifest + steering rows + eval queries | "The five pillars" is presented as the Well-Architected Framework's full set ("measure how far it complies with the AWS Well-Architected Framework"). The WAF has six pillars; Sustainability appears nowhere in the 10,349 added lines: no disclosure, no na-gate, not in the report's Method section. Because the skill discloses its cost-hygiene scope so carefully, the silently dropped pillar reads as totality. | One disclosure sentence ("covers five of the six pillars; Sustainability is not deterministically observable from cluster state") in the description and the report Method section. |
| 16 | `SKILL.md:16-27,250` vs `:253-254,272-310` | The determinism guarantee is over-broad: it holds for the jq-to-HTML path only. The chat headline and the entire markdown deliverable (including hand-filled Rating/Risk columns) are agent-transcribed, and governance answers are interview-transcribed. Nothing tells a reader which surface is machine-vouched. | Scope the claim to the reducer/HTML path explicitly; have the reducer emit band/risk labels, or label the transcribed surfaces. |
| 17 | `SKILL.md:3` + `render-report.py:1606,1616,1847` | Branded "Well-Architected Review" / "complies", with no distinction from the official AWS WA Tool review process and no note that no official EKS Well-Architected Lens exists. Leadership can read the HTML as an AWS-sanctioned WAR outcome. | Add an "unofficial; not the AWS Well-Architected Tool; no official EKS Lens exists" line in SKILL.md and the report Method section. |
| 18 | `render-report.py:1800-1802` + `SKILL.md:264-289` | Caveat asymmetry: only Cost carries a "what this score doesn't mean" caveat. Security/Reliability and the rest are point-in-time config-presence scores with none, so they read as literal health or compliance verdicts. | Add a general point-in-time, config-not-behavior caveat beside the bands. |

## Must fix: high-impact imprecision

| # | Where | Problem | Suggested fix |
|---|---|---|---|
| 19 | `references/reliability.md:53,637,647` (+ `render-report.py:707-709`) | lens-15 detects "private subnets" via `MapPublicIpOnLaunch` only; AWS defines public vs private by route table (IGW route). `routetables.json` is collected and plumbed to the renderer but ignored. False-PASS direction on a weight-3 question, and the scope is cluster-registered subnets while the remediation says "node subnets". The in-code comment defending the proxy ("route-table inspection disagreed 3/3") is consistent with the main-route-table-fallback gotcha: implicitly associated subnets return empty per-subnet association lists, so a naive lookup misses them. | Use route tables with a main-table fallback; align the scope and remediation wording. |

## Should fix

| # | Where | Issue |
|---|---|---|
| 20 | `SKILL.md:152-204` vs `render-report.py:104` | `scores.json` is consumed but no documented step writes it: Step 7 prints to stdout, and the renderer hard-exits without the file (verified both ways by execution). A competent agent infers the redirect, but the improvised step sits on the determinism spine and raises the reachability of finding 23. Add `> "$WORK/scores.json"` to Step 7. |
| 21 | `SKILL.md:11` + references | No read-only hard rule, while `allowed-tools` pre-authorizes all aws/kubectl (including delete verbs) and the references ship executable write remediations (`kubectl delete pv`, `kubectl apply -f` on raw URLs). Sibling eks-upgrade-check carries an explicit "Read-only / assessment-only, hard rule". Mirror it, and state that remediation commands are report content only. |
| 22 | `render-report.py:1130,998-999` | Dark-theme High badge fails WCAG AA: #f9f9fa on #fe6e73 is 2.59:1 at 12px/700 (needs 4.5:1), and it is reachable on every failing High row. (The Critical-badge pair is dead CSS, never rendered.) |
| 23 | `render-report.py:1641,1662,1767` | Three scores.json fields interpolate into HTML unescaped (overall score span, coverage cell, governance heading); a raw `<script>` executes. All cluster-data paths are correctly escaped, so this is a consistency fix: wrap them in `e()`. |
| 24 | `render-report.py` + `SKILL.md:212-213` | The emailable report embeds account-scoped identifiers beyond the reviewed cluster (unrelated OIDC provider ARNs, up to 10 non-cluster volume IDs, VPC endpoint service names) with no masking note. Sibling convention exists (eks-operation-review: "mask or omit the AWS account ID"). Add the masking/disclosure note and consider trimming out-of-scope identifiers. |
| 25 | `references/cost-optimization.md` (lens-12/13) | ECR scan-on-push and immutable tags are scored under Cost Optimization; WAF-wise these are Security controls (the skill itself labels them "security-adjacent"). Move or annotate. |
| 26 | 21 questions across 4 pillar files | Per-question "Detection: ASK USER" tags contradict the authoritative scorers, patched only by a repeated "scorer is authoritative" banner, and the renderer scrapes these sections for report prose. Fix the tags rather than banner-patching. |
| 27 | `misc/evals/.../README.md:31` vs `:46` | Self-contradiction: tasks "need no live cluster" vs `make task-...` "(LIVE, needs a cluster)". Align with the siblings' plain "(LIVE)". |
| 28 | `severity.md:189` / `SKILL.md:3` / `severity.md:23` | Three self-inconsistencies: gp3 "equal/better performance" stated unconditionally while the skill's own cost-analysis.md:170-177 documents the >=1 TiB IOPS downgrade; the description promises "naming the resources behind each finding" while the body admits no-list cases; sec-6's severity rationale is still OIDC-only while the scorer correctly handles Pod Identity. |

## Fargate label assumptions: settled live (one correct, one bug)

AWS docs neither confirm nor deny the two auto-applied-label assumptions the skill relies on (as of 2026-09-04), so we ran the check on a live EKS 1.34 cluster with an active Fargate profile (4 Fargate nodes; coredns and karpenter pods scheduled on them):

- **Nodes: the skill is correct.** Fargate nodes carry `"eks.amazonaws.com/compute-type": "fargate"`, so the mode detection (ope-7/8/15, net-3, `SKILL.md:85`) is fine as shipped.
- **Pods: confirmed bug, filed as finding 29 above.** Fargate pods carry no `compute-type` label at all; their only auto-applied label is `"eks.amazonaws.com/fargate-profile": "default"` (verified on both coredns and karpenter pods; the full label sets contain only component/app labels, `pod-template-hash`, and `fargate-profile`).

## Advisory (fine to defer or handle post-merge)

- Remediation hygiene: the NPD manifest URL 404s (correct file: `deployment/node-problem-detector.yaml`); the `sigstore/cosign-webhook` chart does not exist (current chart: policy-controller); 4 manifest URLs pinned to master/main raw.githubusercontent; cost-6's PV-delete lacks the snapshot-first caution that cost-8 and cost-analysis themselves require; "single biggest EKS attack vector" needs a citation or softening; ASCP name; cloudscape.aws.dev 403s (site is cloudscape.design) and the base font is Open Sans, not Amazon Ember; "IRSA required for cross-account" is stale now that Pod Identity supports target roles (live-verified).
- Rubric texture: several double-penalty pairs (ope-5 + rel-13, sec-27 + sec-28, sec-6 + sec-18 in the partial-IRSA case, one orphaned volume hitting cost-6 + cost-8); the `most` band is unreachable for t=3 questions (ope-2/ope-16) and absent from lens-14's ladder; a mostly-na pillar can render 100/Excellent at exactly 50% coverage with no per-pillar caveat; no eval pins a numeric score, so the determinism property is never eval-tested; severity.md's High/Medium tables list governance-only IDs (for example rel-12 "backups" as High) that never move a number.
- Workflow texture: helper functions and COLLECT_ERRORS assume a single shell and do not survive separate Bash calls (suggest "write to a script and source it"); re-running a pillar scorer appends duplicate JSONL lines with no dedup-by-id in the reducer; the report carries no collection timestamp; Step 8's renderer invocation is cwd-sensitive; the renderer raw-tracebacks on malformed-line or wrong-shape input; "pillar detail below is still valid" renders even when no pillar detail exists; sec-1/sec-2 emit worst-case `none` (false High-class findings) if cluster.json is missing `resourcesVpcConfig`; net-1 falls back to counting all VPC subnets when `subnetIds` is missing.
- Sibling routing polish: eks-recon's judgment clause and eks-cost-intelligence's Distinct-from list omit this skill; there is no eks-security compliance near-miss negative; neg-22's README rationale mislabels an undeclared exclusion; the "WAFR" positive relies on model expansion.
- Architecture note: the severity map lives in three places (SKILL.md jq reducer, renderer SEV sets, severity.md prose) with only a comment as the sync mechanism, referencing a `reduce.sh` that is not in the tree; the reducer looks like it was once a committed asset and belongs back in `assets/` (this is also the root cause of finding 20). Roughly 12 unguarded hardcoded-value clusters (extended-support $/hr, gp2/gp3 prices, the perf-3 regex, tool-name allowlists, raw-master manifest URLs) where the skill's own cost-analysis.md:30 establishes the right "query the Price List API, don't quote fixed numbers" pattern. Detections match tooling by Deployment-name regex while the skill's own cost-analysis documents a real capture where exactly that heuristic misidentified Karpenter as cluster-autoscaler and prescribes image/label matching.

## Pre-existing repo issues surfaced along the way (not this PR's scope, we will track separately)

- The generated skills/README References table drops nested `references/security/*.md` (the generator uses `find -maxdepth 1`), so this skill's largest pillar would be uncatalogued.
- `check_hygiene` failures for eks-mcp-server, eks-recon, eks-upgrade-check pre-exist at base and are not CI-wired.
- The eval runner drops the `expected_output` field framework-wide.

## What verified well

Worth stating plainly, because a lot survived hostile checking:

- The CI surface is genuinely clean: frontmatter validates (description at 1021/1024), all generated files regenerate byte-identically, the website twin is complete, all JSON/YAML parses, and no leaked identifiers anywhere in the branch history.
- The high-consequence numbers all verified exactly right against live sources: the full EKS version/support-date calendar, extended-support pricing and math, AL2 EOL facts, AL2023 migration specifics, gp2/gp3 tables, and the Graviton price ratios (the Price List API reproduced the ap-southeast-5 rates to the cent).
- The measured-vs-governance two-track design is real WAR practice done well; the fail-loud collection helpers and the validation gate genuinely refuse to score denied or empty collections; the renderer output is offline, stdlib-only, byte-identical across Python 3.9/3.13, and escapes cluster data correctly; the band edges are internally consistent across all four restatements.

Happy to re-review as soon as the fixes land; the re-review will again be whole-branch against the new tip.

utkarpun:
Update on N1/N2: no need to run the check, we settled both on a live EKS 1.34 cluster with an active Fargate profile (4 Fargate nodes, coredns and karpenter pods scheduled on them). Split verdict:

**N2 confirmed correct.** Fargate **nodes** do carry the label the skill expects:

```json
"eks.amazonaws.com/compute-type": "fargate"
```

So the mode detection (ope-7/8/15, net-3, SKILL.md:85) is fine as shipped.

**N1 confirmed as a bug, promoting it to must-fix #20.** Fargate **pods** do NOT carry `eks.amazonaws.com/compute-type` at all. Their auto-applied label is:

```json
"eks.amazonaws.com/fargate-profile": "default"
```

(verified on both coredns and karpenter pods; full label sets contain only component/app labels, `pod-template-hash`, and `fargate-profile`). So fargate-2 and fargate-4, which select pods by `eks.amazonaws.com/compute-type=fargate`, match zero pods on a real Fargate cluster and silently no-op to `na`.

Fix: select pods by presence of the `eks.amazonaws.com/fargate-profile` label key (any value), or resolve via the pod's `nodeName` to a node carrying `compute-type=fargate`. Note this interacts with #14: once pod selection works, fargate-4 still needs the `aws-observability/aws-logging` ConfigMap check rather than the fluent-sidecar heuristic.

utkarpun (changes requested):
## Adversarial Review (12-lens): CHANGES_REQUESTED — round 2, whole-branch tip `bee1900`

Thanks Ter Jing — this is a fresh whole-branch re-review (`21ef66d..bee1900`), not a delta. **27 of 29 round-1 findings verified genuinely fixed** (re-executed / re-verified against live AWS docs), zero regressions, zero over-corrections. Strong fix round. What remains is **3 must-fix** (one newly surfaced, two incomplete fixes) plus should-fix/advisory items. CI is clean; none of the must-fix items are CI-caught.

### Must fix

| # | Where | Problem | Fix | Source |
|---|-------|---------|-----|--------|
| 1 | `references/operational-excellence.md:523,527,542` (fargate-4 remediation; surfaced to the customer report via `render-report.py` `question_prose` :170-175/1621) | **Wrong remediation shipped.** Round-1 finding 14 fixed the *scorer* (now reads the `aws-observability`/`aws-logging` ConfigMap at :63-68), but the **remediation prose was left stale** — it still says "Deploy Fluent Bit as a sidecar … `amazon/aws-for-fluent-bit` with FireLens." EKS-on-Fargate uses a **built-in** log router (no sidecar); **FireLens is ECS-only.** A non-passing fargate-4 prints this wrong fix in the report. | Rewrite the remediation to the `aws-observability/aws-logging` ConfigMap path, matching the scorer. | docs.aws.amazon.com/eks/latest/userguide/fargate-logging.html |
| 2 | `SKILL.md` Step 4 + `assets/reduce.sh` + `render-report.py:1741-1746` | **Round-1 finding 11 fix is bypassable — a dead/empty cluster still scores.** Verified by execution at tip: (a) **0 nodes + Pending pods** escapes *both* gates (`NOT_VIABLE` needs `workload_pods==0`; `no_node_ready` needs `node_count>0`) → hero publishes a numeric **100**; (b) the liveness withhold is **renderer-only** — `reduce.sh` emits `technical_overall:100` for an all-`Ready:False` cluster, and SKILL.md Step 7 (`jq -r '.technical_overall'`) + the markdown/chat path report that number before the HTML withhold applies. | Gate viability on `nodes==0 OR (all pods non-Running)`; move the liveness withhold into `reduce.sh` (or the score contract) so every surface — not just HTML — honors it. | — |
| 3 | `render-report.py:1978-1981` (interactive governance) | **Round-1 finding 16 residual (misleading).** Auto-mode governance is correctly labeled "not guessed / separate," but **interactive** governance renders an "X/100 · Y of Z answered" score hero inside a report branded "Deterministic review," with **no label that these are self-reported interview answers**, not machine-measured. A forwarded report reads the governance number as machine-derived. | Add a one-line "self-reported answers, not measured from cluster data" note to the interactive governance body. | — |

### Should fix (IMPRECISE / NUANCED / Medium — not blocking)

| # | Where | Issue |
|---|-------|-------|
| 4 | `render-report.py` exec-summary + hero | Framing weight: the "Risk" column shows bare Low/Med/High derived purely from the config score (`risk()` :82-87) → reads as an assessed threat level; rename **"Risk (from score)"**. And the config-not-behavior/unofficial caveat is `small muted` grey against a 42px score hero — promote it to a visible chip beside the score. |
| 5 | `render-report.py:1857-1866` + `reduce.sh` | A pillar at the gate-minimum **50% coverage** renders a top band ("Excellent · Low risk") with no per-pillar low-coverage caveat (the inline `1/2` cell is the only signal). Add a per-pillar low-coverage marker. |
| 6 | `references/security/identity-access.md:97` (net-2) | Least-privilege SG scorer tests only `.FromPort`, never `.ToPort` — a world-open `80-65535` rule scores "clean". Impact is bounded (render's stricter `_sg_clean` catches it as a loud non-zero exit via the agreement gate, so no silent false-pass), but the detector is still wrong. Test the full port span. |
| 7 | `SKILL.md:10,35-38` + all `references/*.md` | Read-only rule + narrowed allowlist landed (round-1 finding 21), but the "remediation is report content, do not run" caveat lives **only** in SKILL.md while Step 5 loads each pillar file independently (zero per-file caveats); and `allowed-tools` still grants unscoped `Bash(rm:*)`, `Bash(ln:*)`, `Write`. Add a per-file caveat line; scope or drop `rm`/`ln`/`Write`. |
| 8 | `references/cost-analysis.md:21-23` | Graviton family allowlist is hardcoded (`m7g…t4g`; "anything else is x86") with no generation-suffix fallback → a future `m9g` is misclassified x86, contradicting the skill's own "don't hardcode, query live" pattern (perf-3). |
| 9 | `misc/evals/eks-well-architected-review/evals.json` | No eval pins a numeric score or band — the headline product (deterministic scoring) is never machine-tested; a threshold regression passes the suite. Add one score/band-pinned case. |
| 10 | eks-operation-review routing | Live trigger collision: the vendored op-review description ("audit, review, health-check, or **score** an EKS cluster") is unedited (correctly — upstream-owned) and "score/review my EKS cluster" plausibly hits both. Only eval negatives were added; consider a positive disambiguation cue in the WAR description. |
| 11 | `references/cost-optimization.md:202` | `kubectl delete pv` names the orphaned EBS volume but not the `kubernetes.io/pv-protection` finalizer wedge (Terminating hang). Add the survivor/finalizer note. |
| 12 | `SKILL.md:246` | Dangling `CONTEXT.md` pointer — the file was dropped in commit `58adb5f`. Delete the pointer or inline the caveat. |
| 13 | `misc/evals/eks-well-architected-review/.skilleval.yaml` | Missing `artifact_type`/`layers`/`regression` that scored siblings carry — the artifact layer is weighted 0.25 but nothing is wired to validate it. |

### NUANCED / Low (cosmetic — batch at leisure)
- `cost-analysis.md:327` "surcharge is $0.60/hr" mislabels the extended-support *total* (surcharge is $0.50/hr; the ~$365/mo delta is correct). · `reliability.md:54` NAT redundancy counts NATs vs AZ count, not per-AZ placement. · Remediation staleness: NPD URL 404s (`operational-excellence.md:573` → real `deployment/node-problem-detector.yaml`); `workload-security.md:244` cosign chart is `policy-controller` not `cosign-webhook`; Kyverno `validationFailureAction`→`failureAction: Enforce` (:102); sec-16 lists removed PodSecurityPolicies (:46). · `severity.md:21` unsourced "single biggest attack vector" superlative. · Graceful-degradation: malformed `results.jsonl` → raw traceback (`render-report.py:124`) and raw `jq` error (`reduce.sh`); `arn.split(":")[3]` IndexError on truncated ARN (:280/1721); `utcfromtimestamp` DeprecationWarning (Py3.12+); dead `.badge-critical`/`.badge-neutral` CSS. · Band-label case "Needs improvement" vs "Needs Improvement". · 4 sibling descriptions don't reciprocally name this skill (eval-level fan-out is complete). · eval README task-line still says "(LIVE, needs a cluster)" contradicting the new no-cluster caveat.

### Architecture concern
The severity map is triplicated (`reduce.sh` `sev()`, `render-report.py` `SEV3`/`SEV1`, `severity.md` prose) — consistent today, comment-synced, drift-prone; make one authoritative and have the others derive/point. Hardcoded stale-prone values (5 prices, Graviton allowlist, 4 moving-ref manifest URLs) sit in the same files that preach "query the Price List API, don't quote fixed numbers."

### Advisory
- The rendered report embeds the cluster's **own** full OIDC provider ARN (account ID inline, `render-report.py:722/729`) in cleartext — this is by-design for verifiability and carries a "mask before sending" banner; out-of-scope providers are already reduced to counts. No code change required, noted for awareness.
- **Pre-existing repo defect (not this PR's scope, tracked separately):** `skills/README.md` omits this skill's 5 nested `references/security/*.md` — the generator uses `find -maxdepth 1`; the README is byte-consistent with the generator, so this PR didn't introduce it, but the largest pillar is currently uncatalogued.

### Coverage / SHA record
Review unit = `merge-base..HEAD` · merge-base `21ef66d` · tip `bee1900` · non-shallow · all 12 lenses RAN · **F10c port-drift gate N/A (verified — no devops-agent port)**. No unverified residue.

### Strengths
Genuinely strong fix round: the list-disagreement gate now truly fires (banner + non-zero exit), HTML escaping is complete, dark-theme badge contrast fixed, perf-3 previous-gen regex now exact, sec-30/sec-2/sec-21/net-4/lens-15 detectors corrected, six-vs-five-pillar + unofficial disclaimers land on four surfaces, identity/endpoint cross-check added, and the Fargate pod-label bug (finding 29) is resolved via nodeName→node-label. 27/29 verified fixed with no regressions.

utkarpun (changes requested):
## Adversarial Review (12-lens): CHANGES_REQUESTED

Round-3, fresh whole-branch re-review at tip `3a397e27`. One confirmed must-fix (a command injection in `score.sh`); all three round-2 must-fixes verified fixed. Everything else is should-fix or advisory.

### Lenses run
| Lens | Status |
|------|--------|
| 1 Structural / CI | RAN |
| 2 Triggering / Routing | RAN |
| 3 Eval Scaffold / Config | RAN |
| 4 Technical Accuracy | RAN |
| 5 Content Completeness | RAN |
| 6 Scope / Boundary / Security | RAN |
| 7 Source Grounding | RAN |
| 8 External Gap Analysis | RAN |
| 9 Skill Architecture Fitness | RAN |
| 10 Change-Ripple / Old-Claim Sweep | RAN (triggering.json + scoring-rule changes) |
| 11 Eval Determinism / Two-Path (empirical) | RAN |
| 12 Mechanical Output / Rendering (empirical) | RAN |

### Coverage / SHA record
review unit = `merge-base..HEAD` · merge-base `21ef66dbb4708f6c4e78aab88224dbccd20a64fc` · tip `3a397e27a2d1fd4261949458c38e08dacfa1ff5c` · non-shallow verified · tip == PR head (no drift during review) · accumulated full-panel coverage of all 57 changed files (skills/eks-well-architected-review/, misc/evals/*, misc/website mirror, references, assets, index/registry files).

### Port-drift gate (F10c)
gate = N/A (verified — no devops-agent port of eks-well-architected-review: absent from `devops-agent/README.md` registry and no port dir). All subsequent fields → dash.

### INCORRECT (must fix)
| # | File | Claim | Fix | Source |
|---|------|-------|-----|--------|
| 1 | `skills/eks-well-architected-review/assets/score.sh:103` (exec `:114`, interp `:104-106`) | **Command injection via the interactive governance note.** The sanitizer `tr -d '"\\'` strips only `"` and `\`; `$(...)` and backticks survive and are interpolated into a double-quoted `emit` line that runs via `bash "$BLK"`, so a note like `` `touch x` `` / `$(cmd)` executes arbitrary shell in a process holding live AWS creds + kubectl context. Reproduced empirically (both `$(...)` and backtick canaries fired). The note is agent-transcribed from interview answers, which the skill's own confused-deputy threat model (`collect.sh:48`) treats as potentially document-sourced/untrusted. | Single-quote the note or pass it via an env/awk variable instead of interpolating into a double-quoted bash literal; or extend the strip to neutralize `$` and `` ` ``. The note is only ever emitted as an inert JSON `detail` string, so it never needs shell evaluation. | Confirmed by empirical reproduction; sanitizer's own comment (`:101-102`) states the intent was to prevent quote-breakout. |

### IMPRECISE / should-fix
| # | File | Issue | Fix |
|---|------|-------|-----|
| 2 | `SKILL.md:447-450` (+ mirror `index.md:447`) | **Dead-cluster residual (the round-2 dead-cluster must-fix core is fixed).** The overall verdict now correctly prints `NOT VIABLE — no data plane` across all surfaces. But the Step-7 chat jq still prints the five `pillar 100 (coverage 100%)` lines unconditionally beneath it, contradicting `:319`/`:453`. `:319`'s stop-instruction keys on `technical_overall`, which doesn't exist until Step-7 runs, so it can't gate the emission. | Gate `.pillars[]` in the jq on `technical_overall|type=="number"`, or label the pillar lines "declared configuration only" the way `:320` requires for NOT HEALTHY. |
| 3 | `assets/render-report.py` (thresholds `:238-239,243`; caveat docstring `:245-248`; chip `:2371`) | **Risk-band caveat + thresholds never rendered.** report.html shows bare `Excellent/Good/Fair…` + a "risk" chip *styled after the WA Tool's High/Medium Risk labels* (implies an assessed threat level), while the "Risk is the score restated, NOT an assessed threat level" caveat and the band thresholds live only in SKILL.md/code. The end-reader (who never sees SKILL.md) is left to read the chip as an assessed risk. | Render the score-restated caveat + band thresholds into the report Method/legend, or soften the WA-Tool-styled chip so it doesn't imply an assessed threat level. |
| 4 | `references/performance-efficiency.md` (scope) + `SKILL.md:40-64` (gap list) | **Undisclosed PERF03 gap.** Performance Efficiency data-management (storage/volume-type-for-performance, IOPS/throughput) is unscored — defensible — but it's omitted from the gap list that bills itself "exhaustive… the boundary of what was checked," so the omission is invisible to the reader. | Add a one-line "not scored: PERF03 data-management / storage performance" entry to the disclosed gap list. |
| 5 | `SKILL.md:188-198` (prereqs) | **Required read access undocumented.** Prereqs only verify `sts get-caller-identity`; the ~15 read actions collection needs (eks:Describe*/List*, ec2:Describe* set, ecr, cloudtrail, iam:ListOpenIDConnectProviders) + the kubectl-get RBAC (clusterroles/bindings, webhooks, aws-auth CM) are never enumerated. Because collection fails loud, a reviewer with a scoped read role aborts the whole review with no documented remedy. | List the required read IAM actions + RBAC in prereqs. |

### NUANCED / advisory (post-merge, non-blocking)
- **Routing symmetry (Lens 2 F1, refuted as must-fix):** op-review's SKILL.md doesn't name WAR in its Not-for list. One-directional exclusion is the repo's established norm (cost-intelligence→op-review, ingress-migration→op-review are also one-directional) and WAF-framed queries lexically self-route to WAR, so this is cosmetic — optionally add "or Well-Architected review (eks-well-architected-review)" to op-review's Not-for for symmetry.
- **Stale-prone hardcoded values (Lens 9.1):** EKS end-of-support calendar table (`cost-analysis.md:407-411`, verified accurate today) + pinned tool semvers; two version-pinned githubusercontent manifest URLs (`operational-excellence.md:679,1224`) will 404 on tag deletion. Doc self-flags for re-check; consider reading `aws eks describe-cluster-versions` live for the EOSL dates.
- **`render-report.py` two-sources-of-truth (Lens 9.2):** the renderer re-derives verdicts the jq scorers already produce (~3471 lines) — drift risk; SEV weight sets verified byte-equal today.
- **Old workload-pod regex residual (Lens 10):** `SKILL.md:291` (+ mirror) keeps the pre-unification `test("^kube-system$|...")|not` snippet vs reduce.sh's unified `^(kube-|amazon-)`; Step-3 orientation only, not the scored path.
- **Unknown/misspelled pillar name silently dropped (Lens 12):** `render-report.py` iterates a fixed PILLARS list, so a typo'd pillar in scores.json vanishes at exit 0 (shape validated, membership not).
- **`datetime.utcfromtimestamp` DeprecationWarning (`render-report.py:299`)** — noisy stderr, breaks on Python 3.15+.
- **`.skilleval.yaml` is the bare template** — no `artifact_type`/`layers`/`regression` like enriched siblings (op-review, recon); falls to harness defaults though it emits an HTML artifact.
- **Comment-only slip (Lens 7):** `performance-efficiency.md:49-52` calls p2/g2 "current generation" (they're retired); not scored, zero output impact.
- **Region-price staleness (Lens 4):** ap-southeast-5 Graviton on-demand rows + $/hr extended-support rates are self-flagged "verified live, re-check" — staleness caveat, not a demonstrated error.

### Repo-level (NOT this PR's scope — post-merge follow-ups against siblings)
- **eks-cost-intelligence Graviton divergence (Lens 8, HIGH conf, WAR is the correct side):** cost-intelligence hardcodes Graviton "20%" in its dollar math (`spot-graviton-adoption.md:121,374`) and uses the "instance family contains 'g'" detection that WAR explicitly documents as defective — tracked as a separate cost-intelligence issue, not this PR.
- **Pre-existing `make hygiene` red** on eks-mcp-server/eks-recon/eks-upgrade-check unattributed sibling maps (hygiene isn't CI-wired; PR edits are append-only) — not a WAR regression.

### Strengths
- Round-2's three must-fixes all verified fixed: FireLens-on-Fargate now correctly prescribes the `aws-logging` ConfigMap in `aws-observability` and explicitly retracts the ECS FireLens prose; the dead-cluster overall verdict is `NOT VIABLE` consistently across chat/markdown/HTML; the interactive governance score is badged "self-reported — not measured" and excluded from the technical score.
- Empirically hardened scoring spine: missing/malformed input all exit non-zero; reduce.sh↔render-report.py SEV weight sets byte-equal; score bands exhaustive/mutually-exclusive/all-reachable; all AA contrast pairs ≥4.5:1 (no alpha compositing); website mirror byte-identical to a fresh regeneration.
- Read-only confirmed (all `describe/list` + `kubectl get -o json`, no mutating verbs); destructive remediations (PV delete, aws-auth/CRB removal, SSH SG removal) enumerate survivors and gate on preconditions.
- Unusually well-grounded reference set: consequential High-severity claims carry inline AWS doc URLs; severity tiers explicitly disclosed as the skill's own editorial weighting, not AWS-published risk.

